PR #17, opened 2025-04-02
Feature/frontend scaffold

XanderVertegaal:
Frontend components for the Annotator tool. Most are still empty; but for the Parser Input forms a suggestion has been made.

NB: I have updated the Node and Python versions in `test.yml` so the workflow runner uses Node 20 and Python 3.11. Not only is this what our servers will have, but it also fixes a 'test run failure' because of packages not being available for Python 3.7. Node 18 will reach end-of-life at the end of this month, so it is best to prevent warnings about that too.

image

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,8 +41,8 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ['3.8']
-        node-version: ['18.x', '20.x']
+        python-version: ['3.11']
+        node-version: ['20.x']
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python ${{ matrix.python-version }}
--- a/frontend/src/app/annotate/annotation-input/knowledge-base-form/knowledge-base-form.component.ts
+++ b/frontend/src/app/annotate/annotation-input/knowledge-base-form/knowledge-base-form.component.ts
@@ -1,10 +1,103 @@
 import { Component } from "@angular/core";
+import { FormArray, FormGroup, Validators, FormControl } from "@angular/forms";
+import { CommonModule } from "@angular/common";
+import { ReactiveFormsModule } from "@angular/forms";
+import { FontAwesomeModule } from "@fortawesome/angular-fontawesome";
+import { faCheck, faPlus, faTrash } from "@fortawesome/free-solid-svg-icons";
+
+const KnowledgeBaseRelationships = [
+    "EQUAL",
+    "NOT_EQUAL",
+    "SUBSET",
+    "SUPERSET",
+] as const;
+
+type KnowledgeBaseRelationship = typeof KnowledgeBaseRelationships[number];
+
+const relationshipDisplayMapping: Record<KnowledgeBaseRelationship, string> = {
+    EQUAL: "is equal to",
+    NOT_EQUAL: "is not equal to",
+    SUBSET: "is a subset of",
+    SUPERSET: "is a superset of",
+};
+
+interface KnowledgeBaseForm {
+    entity1: FormControl<string>;
+    relationship: FormControl<KnowledgeBaseRelationship>;
+    entity2: FormControl<string>;
+}
 
 @Component({
     selector: "la-knowledge-base-form",
     standalone: true,
-    imports: [],
+    imports: [CommonModule, ReactiveFormsModule, FontAwesomeModule],
     templateUrl: "./knowledge-base-form.component.html",
-    styleUrl: "./knowledge-base-form.component.scss",
+    styleUrls: ["./knowledge-base-form.component.scss"],
 })
-export class KnowledgeBaseFormComponent {}
+export class KnowledgeBaseFormComponent {
+    public form = new FormGroup({
+        kbItems: new FormArray([
+            new FormGroup<KnowledgeBaseForm>({
+                entity1: new FormControl<string>("", {
+                    validators: [Validators.required],
+                    nonNullable: true,
+                }),
+                relationship: new FormControl<KnowledgeBaseRelationship>(
+                    "EQUAL",
+                    {
+                        validators: [Validators.required],
+                        nonNullable: true,
+                    }
+                ),
+                entity2: new FormControl<string>("", {
+                    validators: Validators.required,
+                    nonNullable: true,
+                }),
+            }),
+        ]),
+    });
+
+    public relationshipTypes = KnowledgeBaseRelationships;
+
+    public faCheck = faCheck;
+    public faPlus = faPlus;
+    public faTrash = faTrash;
+
+    constructor() {}
+
+    public addKnowledgeBaseItem(): void {
+        const newItem = new FormGroup<KnowledgeBaseForm>({
+            entity1: new FormControl<string>("", {
+                validators: [Validators.required],
+                nonNullable: true,
+            }),
+            relationship: new FormControl<KnowledgeBaseRelationship>("EQUAL", {
+                validators: [Validators.required],
+                nonNullable: true,
+            }),
+            entity2: new FormControl<string>("", {
+                validators: [Validators.required],
+                nonNullable: true,
+            }),
+        });
+        this.form.controls.kbItems.push(newItem);
+    }
+
+    public getRelationshipTypeName(
+        relationship: KnowledgeBaseRelationship
+    ): string {
+        return relationshipDisplayMapping[relationship];
+    }
+
+    public removeKnowledgeBaseItem(index: number): void {
+        if (this.form.controls.kbItems.length > 1) {
+            this.form.controls.kbItems.removeAt(index);
+        }
+    }
+
+    public onSubmit(): void {
+        if (this.form.valid) {
+            console.log("Submitting:", this.form.value);
+        }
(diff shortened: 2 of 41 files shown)


.github/workflows/test.yml
```diff
@@ -41,8 +41,8 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ['3.8']
-        node-version: ['18.x', '20.x']
+        python-version: ['3.11']
```
bbonf, line 44: nice! maybe time to update the cookiecutter repo as well

frontend/src/app/annotate/annotation-input/knowledge-base-form/knowledge-base-form.component.ts
```diff
@@ -0,0 +1,103 @@
...
+import { faCheck, faPlus, faTrash } from "@fortawesome/free-solid-svg-icons";
+
+const KnowledgeBaseRelationships = [
+    "EQUAL",
+    "NOT_EQUAL",
+    "SUBSET",
+    "SUPERSET",
+] as const;
+
+type KnowledgeBaseRelationship = typeof KnowledgeBaseRelationships[number];
```
bbonf: This is too advanced TypeScript for me, but how does it compare to `type foo = 'qwe' | 'asd' | 'zxc'` ?

frontend/src/app/annotate/annotation-input/knowledge-base-form/knowledge-base-form.component.ts
```diff
@@ -0,0 +1,103 @@
...
+
+@Component({
+    selector: "la-knowledge-base-form",
+    standalone: true,
+    imports: [CommonModule, ReactiveFormsModule, FontAwesomeModule],
+    templateUrl: "./knowledge-base-form.component.html",
+    styleUrls: ["./knowledge-base-form.component.scss"],
+})
+export class KnowledgeBaseFormComponent {
+    public form = new FormGroup({
```
bbonf, line 36: How do these programmatic forms compare to regular bindings?

frontend/src/app/annotate/annotation-input/knowledge-base-form/knowledge-base-form.component.ts
```diff
@@ -0,0 +1,103 @@
...
+                }),
+                relationship: new FormControl<KnowledgeBaseRelationship>(
+                    "EQUAL",
+                    {
+                        validators: [Validators.required],
+                        nonNullable: true,
+                    }
+                ),
+                entity2: new FormControl<string>("", {
+                    validators: Validators.required,
```
bbonf: shouldn't this validator be in a list?

frontend/src/app/annotate/annotation-input/knowledge-base-form/knowledge-base-form.component.ts
```diff
@@ -0,0 +1,103 @@
...
+    public relationshipTypes = KnowledgeBaseRelationships;
+
+    public faCheck = faCheck;
+    public faPlus = faPlus;
+    public faTrash = faTrash;
+
+    constructor() {}
+
+    public addKnowledgeBaseItem(): void {
+        const newItem = new FormGroup<KnowledgeBaseForm>({
```
bbonf, line 77: perhaps the instantiation of a FormGroup can be refactored into a method that takes default values as arguments

bbonf (commented):
Sorry this took so long, I missed this PR completely.

This looks like a good basis. I personally think the input form should maybe become a separate pane that's always visible (say on the left side) and not part of the tab group, so that you could see the effect of changing the input on the parse/proof trees without switching back and forth.

.github/workflows/test.yml
XanderVertegaal: Good idea! I'll contact the System Team to ask about the version we usually get with new CookieCutter servers and get a PR going.

frontend/src/app/annotate/annotation-input/knowledge-base-form/knowledge-base-form.component.ts
XanderVertegaal: `type foo = 'qwe' | 'asd' | 'zxc'` gives me the correct type, but I want the type _and_ an actual array with (only) these values, so I can use it in the template. Since I cannot derive an array from a type, I must derive a type from the array, which is what the `as const` and `[number]` do.

frontend/src/app/annotate/annotation-input/knowledge-base-form/knowledge-base-form.component.ts
XanderVertegaal: Both arrays and single functions are fine, but I agree consistency is best!

frontend/src/app/annotate/annotation-input/knowledge-base-form/knowledge-base-form.component.ts
bbonf: Wouldn't an `Enum` work for that?
(I'm fine with the current solution if you add a comment, just trying to learn more)

frontend/src/app/annotate/annotation-input/knowledge-base-form/knowledge-base-form.component.ts
XanderVertegaal: You mean compared to template-driven forms, with the 'banana box' [(ngModel)]="..." syntax?

I prefer working with these reactive forms (FormGroup, FormControl etc.) because you separate your form from the template by defining it explicitly as such in the component. You can easily get a form control's current value, disable/enable controls and add/remove validators without reaching for the template first, since everything is already defined in your component logic. Another benefit of reactive forms (I think) is that it is easy to subscribe to form changes with form.valueChanges and debounce/filter/map the form values without having to use ngOnChanges or something like that.

frontend/src/app/annotate/annotation-input/knowledge-base-form/knowledge-base-form.component.ts
bbonf: https://www.typescriptlang.org/play/?#code/PTAEFpK6dv4YpzYCh0FMB2BXAtqAMoAuOAZmaAN6qh2gAqAEgJIByA4qALygDkASwDOoAIahiACwFYA5nwA0tegHlmAUQBKPfsLFYA9lIwAnCdLl9UAXwDc6VGRxYAxsQEGs+w85cYAFABuogA2OBgAXESkFACU1Mp0Lp5CBiEYAHQhBrL+gsT8oADUoMFhGLH21uiiWD6uASTkZBnM7ByVqCoARgBWGG4ZANYYAJ5C-k1xGXiiAA7+I6M8AHygyVip6Vk5eVOUSxGKoEuxnaggECjXN7d38A7Y+NHNmhjZogAmGJ8J9IysThKf5qJhaGz2VAbIQFfZvD7fT4AEQwQhcJgEc3cniib2SJk+AB44e8DF8fgpQDCMXI1rwaP8ANokhE-VqAjgAXSighE4ikMnkwPozJiZHhZMRGVBWm5uj5hmMZgFlgh6CcrmxXlq9T8EvJnyCoXCURZkp+8QZ9GhaUy2Vy+UKJTNBpRaIxWI8WEZZXCnM61VQOoMvgw+sRkzF4bZbU4nR6-UGSwmLsRsRm80WY1W6xStp2Dv2JzGR0pp0qQA

XanderVertegaal: An enum would work perfectly fine to this end as well, and it is probably more idiomatic than using `as const` here. I've applied it here! 👍

XanderVertegaal: I have run into problems with enums in the past since they compile to JS in unexpected ways, leading to unwanted types for `Object.values(...)` (if you want to get an array of enum values). However, it seems that this is only true for _numeric_ enums, whereas _string_ enums behave as you would expect. I'm demonstrating the difference here:

https://www.typescriptlang.org/play/?#code/KYOwrgtgBAcpwCcCWBjAwgewDYYVA3gFBQlQBKAogCIA0xpAQgDICqFdpUA4pRTIQF9ChFBhABnAC5RwERKkw4E4qAF4oAeQBGAK2ApJAOgBuAQyxhg4gBRw5ydNlwBKANzCA9B6gAVABZIKpIAngAOwFCBUGAgwAAe4QbAACaGhCHhUACCalAZwBgAZjLwDoq44u5epAB6APzChF7NLa1t7R2dHsKgkFAAypLIIADm5XhEnJRUuQBECCmzHIysFHNaFsBL9CQ8FHxzIwugs4LCohLSUsNjTsq52noGJuaWNoM3426e3v5R+ZEVBoANJpAEMXL5IpQa5IUbjSpNbwkeqEIA

XanderVertegaal: (In all fairness, in this case I did not make an active decision _not_ to use enums; I simply forgot about them / did not consider them, so it is great that you mention them 😄 .)

PR #21, opened 2025-05-22
Feature/problems in db

XanderVertegaal:
Allows the import of SICK and FraCaS datasets and removes the 'example' app from the frontend and backend code.


backend/problem/utils.py
```diff
@@ -0,0 +1,30 @@
+def progress(iteration, total, width=80, start="\r", newline_on_complete=True):
+    """
+    Taken from DIAPP. Creates a progress bar in the console.
```
bbonf: fyi: https://tqdm.github.io/

backend/problem/models.py
bbonf, line 1: While I see some benefits in writing models like that, I'm not sure I completely agree with this approach.

Since SICK and FraCas are static and bounded dataset, I don't think we gain much from modelling them so closely. I think a more generic Problem model would perhaps be neater.
(And could also allow for adding custom problem sets in the future, if this ever becomes relevant)

So maybe the problems can be stored in the db in XML/JSON, and later parsed in python classes that are not db models.

backend/problem/management/commands/import_sick.py
```diff
@@ -0,0 +1,55 @@
...
+
+        existing_sick_problems = get_sick_problems()
+        existing_pair_ids = {p.pair_id for p in existing_sick_problems}
+
+        with open(sick_path, "r", encoding="utf-8") as file:
+            reader = csv.DictReader(file, delimiter="\t")
+            problem_list = list(reader)
+
+            for problem in tqdm(problem_list, desc="Importing SICK problems"):
+                if problem["pair_ID"] in existing_pair_ids:
```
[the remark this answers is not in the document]
XanderVertegaal: This is not true: `problem["pair_ID"]` is a string.

XanderVertegaal:
Round two! How about this architecture?

backend/problem/services.py
XanderVertegaal, line 1: The name `services.py` is for lack of a better alternative. (I've asked Julian and Berit, but we did not really find a good place to put this. 😅)

backend/problem/services.py
```diff
@@ -0,0 +1,70 @@
...
+            problem = SickProblem(
+                    pair_id=problem_data["pair_ID"],
+                    sentence_one=problem_data["sentence_A"],
+                    sentence_two=problem_data["sentence_B"],
+                    entailment_label=problem_data["entailment_label"],
+                    relatedness_score=float(problem_data["relatedness_score"]),
+                )
+            sick_problems.append(problem)
+        except json.JSONDecodeError:
+            print(
```
bbonf: maybe replace `print` with logger?

bbonf: There's an argument to be made about placing a `continue` here, so that the query doesn't fail completely if there's somehow a malformed problem in the DB

backend/requirements.txt
bbonf, line 1: a bit of a silly comment, but if you rerun `pip-compile` without `.\requirements.in` then we could skip these changes

backend/problem/management/commands/import_fracas.py
```diff
@@ -0,0 +1,126 @@
...
+
+            section = problem.get("section")
+            subsection = problem.get("subsection")
+            fracas_answer = problem.get("fracas_answer")
+            fracas_nonstandard = problem.get("fracas_nonstandard", False) == "true"
+
+            premise_nodes = problem.findall("p")
+            premises = [node.text.strip() for node in premise_nodes if node.text]
+
+            with transaction.atomic():
```
bbonf: `atmoic()` is no longer needed

backend/requirements.txt
XanderVertegaal: Haha, no you're right! Still, with `tqdm` (good shout, btw),added, there is an actual change here now.

backend/problem/management/commands/import_fracas.py
XanderVertegaal: You're right! Good to know 👍

PR #22, opened 2025-05-23
Upgrade to Angular 19

XanderVertegaal:
(Drafted until #21 is ready.)

Angular 18 has new, cleaner template syntax, and since the frontend isn't that involved yet, I thought this might be as good a time as any to upgrade, and if we are doing that anyway, we might as well power through to Angular 19.

I've also disabled SSR and removed NgRx (and its associated packages). I don't know why this is in the CookieCutter. We certainly don't need it and it needlessly complicated the application.

This PR should not introduce any functional changes.

If you happen to be working on a frontend feature, I'd be happy to wait until that is merged and solve any merge conflicts.


frontend/package.json
XanderVertegaal, line 1: The linting obscures actual changes. Important packages updated are:
- `angular` (.../animations, .../common, .../core) etc.: ^17.3.0 => ^19.2.13
- `typescript`: ~5.4.2 -> ~5.8.3
- `ng-bootstrap`: ^16.0.0 -> ^18.0.0
- `angular-fontawesome`: ^0.14.1 -> ^1.0.0

frontend/angular.json
```diff
@@ -128,26 +121,22 @@
...
             "i18n": {
                 "sourceLocale": {
-                    "code": "en",
-                    "baseHref": "/en/"
+                    "code": "en"
                 },
                 "locales": {
                     "nl": {
-                        "translation": "locale/messages.nl.xlf",
-                        "baseHref": "/nl/"
```
XanderVertegaal, line 147: Leaving `baseHref` here generates a warning:

```
'baseHref' in 'i18n.locales.nl' may lead to undefined behavior when used with SSR. Consider using 'subPath' instead.

Note: 'subPath' specifies the URL segment for the locale, serving as both the HTML base HREF and the output directory name.
By default, if not explicitly set, 'subPath' defaults to the locale code.
```

backend/langpro_annotator/index.py
```diff
@@ -10,6 +10,9 @@ def index(request: HttpRequest):
     language = request.LANGUAGE_CODE
     page = request.path[1:].split("/", 1)[0]
     # pre-rendered version available?
+    print('Location:', path.join(language, page, "index.html"))
```
bbonf, line 13: If you want to keep those print statements, maybe add an `if settings.DEBUG`

bbonf (commented):
You probably meant to merge this to develop, now that freature/problems-in-db is merged

backend/langpro_annotator/index.py
XanderVertegaal: No, these are simply for debugging purposes to find out why the tests are failing. I cannot run the functional tests locally on my machine, so I have to debug on GitHub (which is not ideal).

If you can run the functional tests on your machine, would you be able to help debugging this?

bbonf: As far as I'm concerned, feel free to rip out the existing functional test code. When we're ready to write tests, I'd like to use playwright (see https://github.com/CentreForDigitalHumanities/ellp for an example)

XanderVertegaal: @Meesch is going to give it his best shot. If not, we will [rip and tear until it is done](https://knowyourmeme.com/memes/rip-and-tear).

PR #31, opened 2025-06-12
Feature/browse problems

XanderVertegaal:
(#22 should be merged first)

In this PR, problems from the SICK and FraCaS datasets are fetched and browsable in the frontend. There are a couple of styling tweaks as well. The two subforms (premises and knowledge bases) are displayed next to one another on screens wider than 1200px and below each other on smaller screens.

image

The SNLI dataset has to be added as well, and the models will have to be updated to handle individual sentences, but I will leave these issues for other PRs.


backend/problem/services.py
```diff
@@ -1,70 +1,117 @@
...
+            pair_id=content["pair_ID"],
+            sentence_one=content["sentence_A"],
+            sentence_two=content["sentence_B"],
+            entailment_label=content["entailment_label"],
+            relatedness_score=float(content["relatedness_score"]),
+        )
+    except json.JSONDecodeError as e:
+        logger.warning(f"Could not decode JSON for Problem ID {instance.id}: {e}")
+        return None
+    except (KeyError, TypeError) as e:
```
bbonf: I'd use `Exception as e` here to be more generic

backend/problem/services.py
```diff
@@ -1,70 +1,117 @@
...
-            print(
-                f"Warning: Could not create SickProblem for Problem ID {problem_obj.id}: {e}"
-            )
-
-    return sick_problems
+    problems = Problem.objects.filter(type=Problem.ProblemType.SICK)
+    return [
+        converted
+        for problem in problems
+        if (converted := instance_to_sick_problem(problem)) is not None
```
bbonf, line 68: TIL you can use `:=` like that

backend/problem/views.py
```diff
@@ -1,3 +1,96 @@
...
+            status=status,
+        )
+
+
+class ProblemView(APIView):
+    def get(self, request, problem_id: int):
+        try:
+            problem = Problem.objects.get(id=problem_id)
+        except Problem.DoesNotExist:
+            return None
```
bbonf: I can't remember how it works, but I think if you don't catch `DoesNotExist` then you'll get a 404 response, which is maybe better than returning None

backend/problem/views.py
```diff
@@ -1,3 +1,96 @@
...
+
+        converted_problem = convert_to_subtype(problem)
+
+        if problem is None:
```
bbonf: if you decide to keep this explicit check, it should happen before `convert_to_subtype`

frontend/src/app/annotate/annotation-input/annotation-input.component.ts
```diff
@@ -1,12 +1,178 @@
...
+            dataset: response.type,
+            judgement,
+            section: response.problem.sectionName,
+            subsection: response.problem.subsectionName,
+            comment: response.problem.note || null,
+        };
+    });
+
+    private getJudgement(problem: ProblemResponse): Judgement {
+        if (!problem.problem) {
```
bbonf: is this just for extra safety or does this mean anything concrete?

frontend/src/app/annotate/annotation-input/problem-details/judgement-badge/judgement-badge.component.ts
```diff
@@ -0,0 +1,44 @@
...
+        switch (judgement) {
+            case Judgement.ENTAILMENT:
+                return "Entailment";
+            case Judgement.CONTRADICTION:
+                return "Contradiction";
+            case Judgement.NEUTRAL:
+                return "Neutral";
+            case Judgement.UNKNOWN:
+            default:
+                return "Unknown";
```
bbonf, line 24: I'd prefer not to have case with a fall-through to default here. Also, I'm not sure why there's a default statement here, I find it a bit confusing.

frontend/src/app/services/annotate.service.ts
```diff
@@ -1,32 +1,51 @@
...
+    // Global database state
+    public firstProblemId: string | null = null;
+    public lastProblemId: string | null = null;
     public totalProblems: number = 1000;
-    public currentProblemIndex: number = 0;
 
-    constructor() {}
+    public problemId$ = new ReplaySubject<string>();
+
+    public problem$ = this.problemId$.pipe(
```
bbonf: Unless there's a good reason to have it this way, I think it'd be nicer to have a method that takes a problem id as an argument. Because now the service has internal state

backend/problem/types.py
```diff
@@ -24,6 +33,19 @@ class FracasProblem:
...
+            "fracasAnswer": self.fracas_answer,
+            "fracasNonStandard": self.fracas_non_standard,
+            "note": self.note,
+            "sectionName": self.section_name,
+            "subsectionName": self.subsection_name,
+            "premises": self.premises,
+        }
+
+
+type CombinedProblem = SickProblem | FracasProblem
```
bbonf, line 51: is this python 3.12 syntax? it's failing for me with python 3.11

bbonf (commented):
Looking good!
Mostly minor comments

backend/problem/services.py
XanderVertegaal: I was really happy to learn this is called the 'walrus operator' by some people. 😂

backend/problem/types.py
XanderVertegaal: I'm afraid it is 😭... I'll replace it with `Union[...]`.

backend/problem/views.py
XanderVertegaal: Ah, no I meant to check `converted_problem` here instead; my bad!

backend/problem/types.py
bbonf: `|` is supported in 3.11, but `type X = ...` isn't.
It seems to work with just `CombinedProblem = SickProblem | FracasProblem`?

frontend/src/app/annotate/annotation-input/annotation-input.component.ts
XanderVertegaal: Just for type safety, as the response may not contain a problem at all. I agree, however, that checking whether the response has a problem should not be the responsibility of this function. I have added a second null check so I can pass a 'real' problem to this function.

backend/problem/types.py
XanderVertegaal: Update: we're sticking with Python 3.12!

frontend/src/app/services/annotate.service.ts
XanderVertegaal: Hmm, I'm not sure I dislike managing state in services that much. For me, the ability to share global data using a service is one of the Angular's most useful features.

However, I do immediately agree that stateless servers are neater (and easier to maintain). I have refactored `problem$` into a function that takes an argument now, with the added benefit that I could make the `annotation-input` component more reactive/declarative 😎.

backend/problem/views.py
XanderVertegaal: According to Copilot (but I can't find the source for this), Django handles this by returning a 500 Internal Server Error, which is even worse than None. It is a good idea to explicitly return a 404 here!

PR #33, opened 2025-06-25
Feature/snli problems

### LINKED ISSUE #29: Add SNLI dataset to DB
Links to the dataset can be found at #27

XanderVertegaal:
Closes #29 

Adds a script to read and add the SNLI problems to the database. The full dataset has 500K problems and will not be imported by default. Only the dev set (10K) is included unless a flag is added.

I have also updated the fetching logic a bit, as I saw that four identical queries are fired every time a new problem is selected. This is fixed now.


backend/problem/models.py
```diff
@@ -1,14 +1,27 @@
...
     )
 
     content = models.JSONField()
+
+    def get_index(self) -> int | None:
+        """
+        Get the index of this Problem in the database.
+        """
+        try:
+            return Problem.objects.filter(id__lte=self.id).count()
```
bbonf, line 24: Why not use `self.id`?

frontend/src/app/services/annotate.service.ts
```diff
@@ -1,34 +1,31 @@
...
-        return this.http.get<ProblemResponse>(`/api/problem/${problemId}`).pipe(
-            catchError(() => {
-                console.log("Error fetching problem");
-                return of(null);
-            })
-        );
-    }
+    public problemId = new Subject<string>();
+
+    public problem$: Observable<ProblemResponse | null> = this.problemId.pipe(
```
bbonf, line 12: Was this intentionally reverted? I don't mind that much, but I'd like to know if you changed your mind

frontend/src/app/annotate/annotation-input/annotation-input.component.ts
```diff
@@ -79,35 +75,44 @@ export class AnnotationInputComponent {
...
-        switch (response.problem.fracasAnswer) {
-            case "yes":
-                return Judgement.ENTAILMENT;
-            case "no":
-                return Judgement.CONTRADICTION;
-            case "unknown":
-                return Judgement.NEUTRAL;
-            case "undefined":
-                return Judgement.UNKNOWN;
+        switch (response.type) {
```
bbonf: Can this switch statement be replaced with a lookup in a dictionary? e.g. `foo[type][label]`
Not important, just more readable IMO

backend/problem/management/commands/import_snli.py
```diff
@@ -0,0 +1,81 @@
...
+        """
+        Import SNLI 1.0 problems from a list of SNLI TSV files and enter them into the database.
+        """
+
+        skipped = 0
+        created = 0
+
+        existing_snli_problems = Problem.objects.filter(type=Problem.ProblemType.SNLI)
+        existing_pair_ids = {
+            json.loads(p.content).get("pairID") for p in existing_snli_problems
```
bbonf: shouldn't `JSONField` do the parsing for you?

backend/problem/management/commands/import_snli.py
```diff
@@ -0,0 +1,81 @@
...
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--full",
+            action="store_true",
+            help="Import the full SNLI dataset (dev, train, and test sets). Warning: this will take a few minutes.",
+        )
+
+    def handle(self, *args, **options):
+        snli_paths = [("dev", "problem/data/snli_1.0_dev.txt")]
+        if options["full"]:
```
bbonf: Nit picking, but I'd make the command line take the file names of the files you want to load, instead of having hardcoded paths/options

backend/problem/management/commands/import_snli.py
```diff
@@ -0,0 +1,81 @@
...
+                        # Handle empty gold labels.
+                        if problem["gold_label"] == "-":
+                            problem["gold_label"] = "none"
+
+                        # Handle empty labels.
+                        for key in ["label1", "label2", "label3", "label4", "label5"]:
+                            if problem[key] == "":
+                                problem[key] = "none"
+
+                        created += 1
```
bbonf: Nit picking, but this would be incorrect if `create()` throws an exception. Should be moved to after `create()`

backend/problem/management/commands/import_snli.py
```diff
@@ -0,0 +1,81 @@
...
+                        Problem.objects.create(
+                            type=Problem.ProblemType.SNLI,
+                            content=json.dumps(problem),
```
bbonf: I think you're not supposed to use `json.dumps()` here because content is a `JSONField`, see comment above. I'm guessing you get doubly-encoded JSON because of that. But I'm not 100% sure.

backend/problem/models.py
bbonf: To elaborate a bit: I think what you're looking for here is the problem's position in the problem set that the user is currently processing. I say this because of the `{{index}} of {{total}}` fragment in the view. However, because the index is not intrinsic to the problem but is affected by the whole set of problems, this could cause some confusion.

frontend/src/app/services/annotate.service.ts
```diff
@@ -1,34 +1,31 @@
...
-            return of(null);
-        }
-        return this.http.get<ProblemResponse>(`/api/problem/${problemId}`).pipe(
-            catchError(() => {
-                console.log("Error fetching problem");
-                return of(null);
-            })
-        );
-    }
+    public problemId = new Subject<string>();
```
bbonf, line 10: Would something like this work?

```suggestion
    public problem$ = new ReplaySubject<ProblemResponse | null>(1);

    public load(problemId: string) {              
        this.http.get<ProblemResponse>(`/api/problem/${problemId}`).pipe(
            catchError(() => {
                console.error("Error fetching problem");
                return of(null);
            }),
        ).subscribe(this.problem$);
    }

```

backend/problem/management/commands/import_snli.py
XanderVertegaal: You're right!

backend/problem/management/commands/import_snli.py
XanderVertegaal: Indeed, and the same is true for SICK and FraCaS. I'm fixing all of them! 👍

backend/problem/management/commands/import_snli.py
XanderVertegaal: It should, and it does! (I've removed the loading/dumping of JSON w.r.t. JSONFields.)

backend/problem/models.py
XanderVertegaal: No, this was actually to solve the issue that I ran into when I cleared the database and ran the import again: the IDs of the records kept increasing (i.e. starting somewhere in the 560000s), so the frontend would show something like "Problem 560012 out of 10000" or something like that.

frontend/src/app/services/annotate.service.ts
XanderVertegaal: Yes, this was intentional. If I recall correctly, duplicate requests were made because both AnnotationInputComponent and NavigatorComponent were sending the current route to AnnotateService. In the current setup, the top component in the annotation module provides the route and AnnotationInput and Navigator simply consume the retrieved problem. It also allows me to move the `switchMap` to the AnnotateService, so overall it feels a bit neater.

frontend/src/app/services/annotate.service.ts
XanderVertegaal: It probably would, but I expect this create new subscriptions on every `load()` call. Moreover, each subscription will stay active for as long as the application is active. Now in this case, `http.get()` automatically completes once data is received, and you could argue that we want to keep listening for new problems for the app's full lifetime anyway, so in practice these points are a bit moot.

In general, however, I try to subscribe only in components so I can complete the subscriptions when the component is destroyed, and to avoid subscribing in services as much as possible. Component-level services (non-singleton, not provided in 'root') are an exception to this, since they are tightly bound to a component and can also implement `onDestroy` to clear up any subscriptions.

PR #35, opened 2025-06-25
Bug/refactor problem details

XanderVertegaal:
(Review after #33 )

A small bug: the 'problem details' (ID, judgement label, dataset etc.) do not show up if the page is loaded for the first time. This is because the template subscribes too late to catch the first emission.

image

I have added a `shareReplay` to the AnnotateService to fix this.

I have also moved code specifically aimed at formatting the ProblemDetails to the ProblemDetailsComponent, so the AnnotateInput component has less bloat.

### DIFF
--- a/frontend/src/app/annotate/annotation-input/annotation-input.component.html
+++ b/frontend/src/app/annotate/annotation-input/annotation-input.component.html
@@ -8,9 +8,7 @@
 @if (form$ | async; as inputForm) {
 <div class="card">
     <div class="card-body">
-        @if (problemDetails$ | async; as details) {
-        <la-problem-details [problemDetails]="details" />
-        }
+        <la-problem-details />
         <div class="input-form-container">
             <la-premises-form class="subform" [form]="inputForm" />
             <la-knowledge-base-form class="subform" [form]="inputForm" />
--- a/frontend/src/app/annotate/annotation-input/annotation-input.component.html
+++ b/frontend/src/app/annotate/annotation-input/annotation-input.component.html
@@ -8,7 +8,7 @@
 @if (form$ | async; as inputForm) {
 <div class="card">
     <div class="card-body">
-        <la-problem-details />
+        <la-problem-details [problem]="problem$ | async" />
         <div class="input-form-container">
             <la-premises-form class="subform" [form]="inputForm" />
             <la-knowledge-base-form class="subform" [form]="inputForm" />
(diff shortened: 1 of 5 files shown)


frontend/src/app/annotate/annotation-input/problem-details/problem-details.component.ts
```diff
@@ -39,4 +54,83 @@ export class ProblemDetailsComponent {
         }
         return null;
     });
+
+    private extractDetails(
```
bbonf, line 76: I know this is not new code, but I'm now wondering whether this fits better in the backend. I don't have a strong opinion

frontend/src/app/annotate/annotation-input/problem-details/problem-details.component.ts
```diff
@@ -39,4 +54,83 @@ export class ProblemDetailsComponent {
...
+                    dataset: response.type,
+                    judgement,
+                    section: null,
+                    subsection: null,
+                    comment: null,
+                };
+        }
+    }
+
+    private getJudgement(response: ProblemResponse): Judgement {
```
bbonf, line 114: same re: backend/frontend

frontend/src/app/annotate/annotation-input/annotation-input.component.html
```diff
@@ -8,9 +8,7 @@ <h2 i18n>Parser input</h2>
 @if (form$ | async; as inputForm) {
 <div class="card">
     <div class="card-body">
-        @if (problemDetails$ | async; as details) {
-        <la-problem-details [problemDetails]="details" />
-        }
+        <la-problem-details />
```
bbonf: I think this should take problem as input

frontend/src/app/annotate/annotation-input/problem-details/problem-details.component.ts
XanderVertegaal: I agree!

frontend/src/app/annotate/annotation-input/annotation-input.component.html
XanderVertegaal: Yes, makes sense!

frontend/src/app/annotate/annotation-input/problem-details/problem-details.component.ts
XanderVertegaal: I agree it might be better. I'll create an issue for this: #35

PR #39, opened 2025-07-08
Feature/problem model updates

XanderVertegaal:
This PR updates the Problem model.

Fields shared by all models (`premises`, `hypothesis`, `entailment_label`) are homogenized and set as model fields. The frontend no longer has to do conversions from dataset-specific entailment labels to a general one. The backend takes care of this.

The `content` JSONField has been renamed to `extra_data` and now contains info specific to the individual problem sets. The methods importing and serializing this problem-specific data are kept close to one another, within the same class.


backend/problem/management/commands/import_snli.py
```diff
@@ -53,8 +61,10 @@ def import_snli_problems(self, snli_paths: list[tuple[str, str]]) -> None:
         skipped = 0
         created = 0
 
-        existing_snli_problems = Problem.objects.filter(type=Problem.ProblemType.SNLI)
-        existing_pair_ids = {p.content.get("pairID") for p in existing_snli_problems}
+        existing_snli_problems = Problem.objects.filter(dataset=Problem.Dataset.SNLI)
+        existing_pair_ids = {
+            p.additional_content.get("pairID") for p in existing_snli_problems
```
bbonf: `additional_content` should be `extra_data`, `pairID` should be `pair_id`

But I actually came here to say that you could get everything in a single query instead:
```
existing_snli_problems.values_list('extra_data__pair_id', flat=True)
```
It really doesn't matter in this context, but it's a small example of how you can use `JSONField` in queries, which can be useful since PostgreSQL is JSON aware and in some cases that's going to be much faster than first parsing everything in python.

See: https://docs.djangoproject.com/en/4.2/topics/db/queries/#querying-jsonfield and https://www.postgresql.org/docs/current/functions-json.html

backend/problem/services.py
```diff
@@ -1,148 +1,17 @@
...
+from langpro_annotator.logger import logger
 
 
 def get_related_problem_ids(problem_id: int) -> Tuple[int, int, int]:
+    pass
     """
     Retrieves the IDs of the next, previous, and random Problem objects
     in the database relative to the given problem ID.
     """
+    from problem.models import Problem
```
bbonf: Why is this import not at the top?

backend/problem/services.py
```diff
@@ -1,148 +1,17 @@
...
-        return instance_to_fracas_problem(problem)
-    elif problem.type == Problem.ProblemType.SNLI:
-        return instance_to_snli_problem(problem)
-    else:
-        return None
+from langpro_annotator.logger import logger
 
 
 def get_related_problem_ids(problem_id: int) -> Tuple[int, int, int]:
+    pass
```
bbonf: Leftover from debugging?

backend/problem/management/commands/import_snli.py
XanderVertegaal: I'm still getting used to this. Really good to know!

backend/problem/services.py
XanderVertegaal: I'm afraid so. Sorry about this!

backend/problem/services.py
XanderVertegaal: This was to 'solve' a circular import, but I'm not happy with it. I'll move this function to another file instead.

bbonf (approved):

PR #41, opened 2025-07-23
Feature/data submission

XanderVertegaal:
Submits problem data to the backend. For existing problems (SNLI/FraCaS/SICK) this may turn out not to be relevant (unless users are allowed to edit them), but we will need this for problems that are wholly designed by the user.

### DIFF
--- a/backend/langpro_annotator/http_client.py
+++ b/backend/langpro_annotator/http_client.py
@@ -1,0 +1,3 @@
+import urllib3
+
+http_client = urllib3.PoolManager()
--- a/backend/langpro_annotator/http_client.py
+++ b/backend/langpro_annotator/http_client.py
@@ -1,3 +1,0 @@
-import urllib3
-
-http_client = urllib3.PoolManager()
--- a/backend/problem/views/parse.py
+++ b/backend/problem/views/parse.py
@@ -68,8 +68,8 @@ class ParseView(APIView):
     def send_to_parser(self, data: ParserInput) -> dict | None:
         """Send frontend data to downstream LangPro service."""
 
-        print("Sending to LangPro service:", asdict(data))
-
+        logger.info("Sending to LangPro service:", asdict(data))
+        # TODO: Implement actual HTTP request to LangPro service.
         # try:
         #     langpro_response = http_client.request(
         #         method="POST",
--- a/backend/problem/views/parse.py
+++ b/backend/problem/views/parse.py
@@ -1,9 +1,10 @@
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, asdict
 
 from django.http import JsonResponse
 from rest_framework.views import APIView
 from rest_framework.request import Request
 
+from problem.types import KnowledgeBase
 from langpro_annotator.logger import logger
 
 
@@ -23,20 +24,6 @@ class ParseResponse:
 
 
 @dataclass
-class KnowledgeBase:
-    entity1: str
-    entity2: str
-    relationship: str
-
-    def to_dict(self) -> dict:
-        return {
-            "entity1": self.entity1,
-            "entity2": self.entity2,
-            "relationship": self.relationship,
-        }
-
-
-@dataclass
 class ParserInput:
     """Represents the input for the LangPro parser."""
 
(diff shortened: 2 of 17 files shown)


backend/problem/views/proof_bank_stats.py
XanderVertegaal, line 1: No functionality change here. I've just moved the two views we already had to separate files.

frontend/src/app/annotate/annotate.component.ts
```diff
@@ -18,21 +14,4 @@ import { takeUntilDestroyed } from "@angular/core/rxjs-interop";
...
-    constructor(
-        private destroyRef: DestroyRef,
-        private route: ActivatedRoute,
-        private annotateService: AnnotateService
-    ) {}
-
-    ngOnInit(): void {
-        this.route.params
-            .pipe(
-                map((params) => params["problemId"]),
```
XanderVertegaal, line 31: Moved this to AnnotateInputComponent.

This triggers the fetching of the Problem, and the AnnotateInputComponent is also the one listening to the results, so it makes sense to have them together. Now AnnotateComponent can be 'dumb'.

frontend/src/app/annotate/annotation-input/annotation-input.component.ts
```diff
@@ -43,50 +43,67 @@ export type AnnotationInputForm = FormGroup<{
...
-
-    public form$ = this.problem$.pipe(
-        map((response) => this.buildForm(response)),
-    );
-
-    private formSignal = toSignal(this.form$, {
-        initialValue: null,
-    });
+export class AnnotationInputComponent implements OnInit {
+    public form: AnnotationInputForm | null = null;
```
XanderVertegaal, line 55: This replaces the older signal-based approach. It turns out that form value changes are not picked up if they are wrapped in a signal. (This is something the Angular team is currently working on, apparently.)

For now, the classic approach of setting a component property in a subscription will do.

backend/problem/views/parse.py
```diff
@@ -0,0 +1,85 @@
...
+                    KnowledgeBase(**kb) for kb in payload.get("kbItems", [])
+                ],
+            )
+
+            response = self.send_to_parser(parser_input)
+
+            return ParseResponse(data=response).json_response(status=200)
+
+        except Exception as e:
+            logger.error(f"An error occurred in ParseView: {e}")
```
bbonf: I'd use `logger.exception()` so that you get the stack trace (not relevant for e.g. Sentry, but may be useful in the console)

backend/langpro_annotator/http_client.py
```diff
@@ -0,0 +1,3 @@
+import urllib3
+
+http_client = urllib3.PoolManager()
```
bbonf: I see this isn't used yet, but I'd choose `requests` just because it's more common/simple. Unless you have a good reason to choose `urllib3`? I've never used it

backend/problem/views/parse.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+@dataclass
+class ParseResponse:
+    error: str | None = None
+    data: dict | None = None
+
+    def json_response(self, status=200) -> JsonResponse:
+        return JsonResponse(
+            {
```
bbonf: Would `asdict(self)` work here?

backend/problem/views/proof_bank_stats.py
```diff
@@ -0,0 +1,37 @@
...
+                "lastProblemId": self.last_problem_id,
+                "totalProblems": self.total_problems,
+            },
+            status=status,
+        )
+
+
+class ProofBankStatsView(APIView):
+    def get(self, request):
+        first_problem = Problem.objects.order_by("id").first()
```
bbonf:
```
qs = Problem.objects.order_by("id")
first_problem = qs.first()
last_probme = qs.last()
total_problems = qs.count()
```
Would be useful when you have to apply filters to the queryset

frontend/src/app/services/annotate.service.ts
```diff
@@ -30,5 +39,16 @@ export class AnnotateService {
             })
         );
 
+    public parse$ = this.submit.pipe(
+        switchMap((form) =>
+            this.http.post<ProblemResponse>("/api/problem/parse", form).pipe(
+                catchError(() => {
+                    console.error("Error parsing problem");
```
bbonf: maybe log the error as well?

backend/problem/views/parse.py
XanderVertegaal: Oh, I didn't know this was a thing! Good call! :)

backend/langpro_annotator/http_client.py
XanderVertegaal: I took this from ParsePort and didn't think too much about it. Looking at `requests` now, it seems to offer a much simpler interface, so I'd be happy to switch to that. 👍

backend/problem/views/parse.py
XanderVertegaal: It does!

frontend/src/app/services/annotate.service.ts
XanderVertegaal: Sure, although I'd like to do some proper error handling in the future (notifying the users as well).

PR #43, opened 2025-07-31
Feature: Sentence and KB models

XanderVertegaal:
Adds KnowledgeBase and Sentence models, and the relationships between them.

The annotation models will follow in a separate PR.


backend/problem/management/commands/import_fracas.py
```diff
@@ -90,23 +92,37 @@ def import_fracas_problems(self, fracas_path: str) -> None:
...
+
+            premises = []
             premise_nodes = problem.findall("p")
-            premises = [node.text.strip() for node in premise_nodes if node.text]
+            for node in premise_nodes:
+                if node.text:
+                    premises.append(Sentence.objects.create(text=node.text.strip()))
+
+            fracas_answer = problem.get("fracas_answer")
+            if fracas_answer is None:
```
bbonf: Perhaps this check should happen before creating the sentences?

backend/problem/migrations/0004_sentence_remove_problem_premises_knowledgebase_and_more.py
bbonf, line 1: This migration may be deleted, as we can start with a fresh import

PR #44, opened 2025-07-31
Feature/annotation models

### LINKED ISSUE #42: Updated data model
New models to enable sharing sentences (with CCG parses) across problems and accommodate future annotations.

Image

Open question: are knowledge base items shared across problems or problem-specific?

XanderVertegaal:
To be reviewed after #43 

Closes #42  

Adds a new Django app 'annotation' along with associated models that we can use in the future. Nothing is being done with these yet.


bbonf (approved):
The chosen approach means we have to replicate all editable fields of the source models in the annotation models. I think that's a good compromise between DRY and KISS 👍

jgonggrijp:
> compromise between DRY and KISS

Just reacting to this particular phrase. Are you just referring to this particular situation, or do you believe that DRY and KISS are generally at odds with each other?

XanderVertegaal:
> > compromise between DRY and KISS
> 
> Just reacting to this particular phrase. Are you just referring to this particular situation, or do you believe that DRY and KISS are generally at odds with each other?

This reminded me once more that regardless of repo or team, there is always a @jgonggrijp reading your comments 🥷

bbonf:
> > compromise between DRY and KISS
> 
> Just reacting to this particular phrase. Are you just referring to this particular situation, or do you believe that DRY and KISS are generally at odds with each other?

Ideally these principles combine with beautiful abstractions etc 🦄🌈.

But in this specific case of reusing (part of) a db model spec in a derived model, while it's certainly doable,  I think it's going to get ugly fast as you work against Django's grain. And what do you stand to gain? alleviating some manual work in case you need to add some fields at a later moment.

jgonggrijp:
> But in this specific case of reusing (part of) a db model spec in a derived model, while it's certainly doable, I think it's going to get ugly fast as you work against Django's grain. And what do you stand to gain? alleviating some manual work in case you need to add some fields at a later moment.

It's not just work but also surface area where things can go wrong (which compounds the more you repeat something). That being said, I can see that replicating the fields may be the best option in this case. It's probably no coincidence that Django forms and admins, DRF serializers etcetera also require some forms of replication.

PR #45, opened 2025-08-06
expand/collapse tree nodes on click

bbonf:


frontend/src/app/annotate/tableau-svg/tableau-tree.component.ts
```diff
@@ -78,4 +89,26 @@ export class TableauTree {
...
+        // the last node can't be collapsed
+        // unless there are subtrees
+        if (idx < this.expanded.length - 1 || this.tree.subtrees) {
+            this.expanded[idx] = !this.expanded[idx];
+        }
+    }
+
+    visibleNodes() {
+        let firstCollpased = this.expanded.indexOf(false);
+        if (firstCollpased == -1) {
```
XanderVertegaal, line 103: This can be a `const`, and there's a typo in the name.

frontend/src/app/annotate/tableau-svg/tableau-term.component.ts
XanderVertegaal, line 1: It's [advisable](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html) to use `string` instead of `String` as a type.

Also, in cases such as `end: boolean = false;`, TS can infer the type from the default value, so a type annotation is technically unnecessary (but doesn't hurt either).

frontend/src/app/annotate/tableau-svg/tableau-tree.component.svg
XanderVertegaal, line 1: The `track` expressions give me console warnings. The one on line two can be fixed by replacing

```
track node.idx
```

with 

```
track `${node.idx}-${$index}`
```

The other two (for the subtrees) below can hopefully be solved in a similar way.

frontend/src/app/annotate/tableau-svg/tableau-tree.component.scss
```diff
@@ -0,0 +1,4 @@
+.expand-plus {
+    font-size: 1.8em;
+    cursor: pointer;
```
XanderVertegaal: Instead of `cursor: pointer`, you can also add `role="button"` to this node (same goes for terms).

XanderVertegaal (approved):
This works great! 🌟 

It would be great if you could solve the console warnings regarding `track`, but this is not a high-priority issue.

frontend/src/app/annotate/tableau-svg/tableau-tree.component.scss
bbonf: I think `role` attributes are not supported in SVG
Edit: but it seems to work, so that's alright

PR #46, opened 2025-08-07
Feature/filter problems

XanderVertegaal:
Introduces a filter form in the frontend. The URL is updated and new requests are made upon changing its value.

The backend does not do anything with these added query parameters yet. I want to save that for a future PR.

image


frontend/src/app/annotate/annotation-input/annotation-input.component.ts
```diff
@@ -45,28 +45,26 @@ export type AnnotationInput = ReturnType<AnnotationInputForm["getRawValue"]>;
...
     public faCheck = faCheck;
 
-    constructor(
-        private route: ActivatedRoute,
-        private destroyRef: DestroyRef,
-        private annotateService: AnnotateService
-    ) {}
-
     ngOnInit(): void {
-        this.annotateService.problem$
```
XanderVertegaal, line 69: I've removed the "AnnotateService" (which did nothing with annotations at all at the moment) and split its contents into  a ProblemService (for fetching/filtering Problems) and a ParseService (for triggering the LangPro parser).

frontend/src/app/annotate/annotation-input/problem-details/problem-details.component.ts
```diff
@@ -37,6 +38,7 @@ export class ProblemDetailsComponent {
     });
 
     public faQuestionCircle = faQuestionCircle;
+    public datasetLabels = datasetLabels;
```
bbonf, line 41: Is this used?

frontend/src/app/annotate/search/search.component.ts
```diff
@@ -0,0 +1,145 @@
...
+        text: new FormControl<string | null>(""),
+    });
+
+    // TODO: Use actual loading state...
+    loading$ = new BehaviorSubject<boolean>(false);
+
+    public faSearch = faSearch;
+    public faTimes = faTimes;
+
+    private entailmentLabels = entailmentLabels;
```
bbonf: Why do these have to become members? why not use the imported variable directly?

frontend/src/app/annotate/search/search.component.ts
```diff
@@ -0,0 +1,145 @@
...
+                takeUntilDestroyed(this.destroyRef)
+            )
+            .subscribe((value: SearchParams) => {
+                this.updateUrl(value);
+            });
+
+        // Update form when URL changes.
+        this.route.queryParamMap.pipe(
+            distinctUntilChanged((prev, curr) => {
+                return prev.get('dataset') === curr.get('dataset') &&
```
bbonf: If I understand correctly, this is needed because there may be query parameters in the URL that are unrelated to this form. But what's the harm of always updating the form based on the URL?

Otherwise, maybe first filter `this.route.queryParamMap` and then there's no need to compare individual values?

frontend/src/app/annotate/search/search.component.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    public clearFilters(): void {
+        this.form.reset();
+    }
+
+    // Updates the route, which triggers a new query.
+    private updateUrl(searchParams: SearchParams): void {
+        const url = this.router.createUrlTree([], {
+            relativeTo: this.route,
+            queryParams: this.formatQueryParams(searchParams)
```
bbonf: What happens if you used `searchParams` directly?

frontend/src/app/services/problem.service.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    private extractSearchParams(routeParams: ParamMap): HttpParams {
+        const text = routeParams.get("text");
+        const dataset = routeParams.get("dataset");
+        const gold = routeParams.get("gold");
+        const entailmentLabel = routeParams.get("entailmentLabel");
+
+        const paramRecord: Record<string, string> = {};
+
+        if (text) {
```
bbonf: `paramRecord = {text, dataset, gold, entailmentLabel}`?

frontend/src/app/annotate/annotation-input/problem-details/problem-details.component.ts
XanderVertegaal: The template uses it, in:

```html
<tr>
    <td class="text-muted" i18n>Dataset:</td>
    <td>{{ datasetLabels[details.dataset] }}</td>
</tr>
```

frontend/src/app/annotate/search/search.component.ts
XanderVertegaal: In this case there is no reason. I think I used them in the template before, but not anymore. Consider them gone :)

frontend/src/app/annotate/search/search.component.ts
XanderVertegaal: Everything we need. The helper function can go.

frontend/src/app/services/problem.service.ts
XanderVertegaal: Not quite, as that would create keys with `null` values. I don't want the value to exist at all in the params if there is no value for it.

frontend/src/app/annotate/search/search.component.ts
XanderVertegaal: I think this was to avoid an infinite loop, but I cannot seem to reproduce that now, so I'm removing the filter and maybe re-add it when stuff breaks.

PR #48, opened 2025-08-13
Feature/filter in backend

### LINKED ISSUE #30: Search problems
Two search functions:

- Full text search (for premises/hypothesis); a styled and functional example can be found in Lettercraft.
- Filters:
  - Judgement / inference label:
  - Dataset
  - Gold status

For instance: "Give me all problems labeled with CONTRADICTION for the SNLI dataset"

XanderVertegaal:
Closes #30 

With this PR, the backend actually filters problems based on parameters sent by the frontend.

I've merged the ProblemView and ProblemDetailView, since the data they provide is requested at the same time anyway.


backend/problem/models.py
```diff
@@ -47,12 +48,12 @@ class EntailmentLabel(models.TextChoices):
...
     extra_data = models.JSONField()
 
-    def get_index(self) -> int | None:
+    def get_index(self, qs: QuerySet) -> int | None:
         """
         Get the index of this Problem in the database.
         """
         try:
-            return Problem.objects.filter(id__lte=self.pk).count()
+            return qs.filter(id__lte=self.pk).count()
```
bbonf, line 56: if the Problems are filtered, then `id < self.pk` is no longer a stable index. But I guess it does what you intend it to do, so maybe the comment has to be updated.

backend/problem/problem_details.py
```diff
@@ -1,28 +1,81 @@
...
+        return None
+
+    filters = Q()
+    if dataset:
+        filters &= Q(dataset=dataset)
+    if entailment_label:
+        filters &= Q(entailment_label=entailment_label)
+    if gold is not None:
+        # To be implemented!
+        pass
```
bbonf: you could use `raise NotImplementedError()` to make it more visible

backend/problem/problem_details.py
```diff
@@ -1,28 +1,81 @@
...
+    """
+    dataset = query_params.get("dataset")
+    entailment_label = query_params.get("entailmentLabel")
+    gold = query_params.get("gold")
+    text = query_params.get("text")
+
+    if not (dataset or entailment_label or gold or text):
+        return None
+
+    filters = Q()
```
bbonf, line 68: nice approach for building a filter

backend/problem/problem_details.py
```diff
@@ -1,28 +1,81 @@
...
+    if dataset:
+        filters &= Q(dataset=dataset)
+    if entailment_label:
+        filters &= Q(entailment_label=entailment_label)
+    if gold is not None:
+        # To be implemented!
+        pass
+    if text:
+        filters &= Q(hypothesis__text__icontains=text) | Q(
```
bbonf: I'd add parentheses around the OR operation for clarity and formatting

frontend/src/app/services/problem.service.ts
```diff
@@ -55,12 +55,13 @@ export class ProblemService {
...
-        .get<ProofBankStats>("/api/problem/proofbank-stats")
-        .pipe(
-            catchError((error) => {
-                console.error(`Error fetching ProofBank stats:`, error);
-                return of(null);
-            })
-        );
+    // Retrieves the ID of the first available problem so we can link to it
+    // from the main navigation bar.
+    public getFirstProblemId$ = this.http.get<{ firstProblemId: string | null; }>("/api/problem/entry").pipe(
```
bbonf: why can't you reuse the generic problem view with no query parameters?

XanderVertegaal: Great idea! 🤩

backend/problem/problem_details.py
```diff
@@ -71,11 +71,10 @@ def get_filters(query_params: QueryDict) -> Q | None:
     if entailment_label:
         filters &= Q(entailment_label=entailment_label)
     if gold is not None:
-        # To be implemented!
-        pass
+        raise NotImplementedError()
     if text:
-        filters &= Q(hypothesis__text__icontains=text) | Q(
-            premises__text__icontains=text
+        filters &= Q(
```
bbonf, line 76: I meant to suggest only a pair of `()`, while you added `Q()`. Maybe this is also fine but I've never nested `Q()`s so I don't know

XanderVertegaal: I know, and I should have put this in a comment, but single `()` gets undone by my editor/linter upon hitting save. Using `Q()` instead provides the same reading benefits as `()` and it doesn't change the function.

PR #50, opened 2025-08-27
prepare for functional tests using playwright

bbonf:


XanderVertegaal (commented):
It's ya boi Xander with a Windows machine 🥸

The yarn command does not work for me unless I install and add [cross-env](https://github.com/kentcdodds/cross-env) as a dev dependency to set the env vars correctly. I also replaced "PYTHONPATH=.:backend" with "PYTHONPATH=./backend". I'm not sure what the `.:` does on Linux.

```suggestion
    "scripts": {
        ...
        "test-func": "cross-env PYTHONPATH=./backend DJANGO_SETTINGS_MODULE=langpro_annotator.settings DJANGO_ALLOW_ASYNC_UNSAFE=1 pytest -k functional-tests",
        ...
    },
    "devDependencies": {
        ...
        "cross-env": "^10.0.0"
    }
```

My next problem is that [`pexpect` does not support `spawn` on Windows](https://pexpect.readthedocs.io/en/stable/overview.html#pexpect-on-windows). I tried installing `wexpect` and replaced `pexpect.spawn` with `wexpect.spawn`; this gets rid of the error, but now the test seems to hang indefinitely. I'm also not sure `wexpect` works on Linux anyway.

We might have to look at this together at some point, or accept that I won't be able to run functional tests. The code looks good, in any case.

XanderVertegaal (approved):
This works! <3 On behalf of all Windows users: thanks for accommodating 😉

(If you could increase the timeout to 120 sec or so, that would be great. 60 seconds is just a little short for my setup, apparently.)

PR #53, opened 2025-09-16
Feature/user problems

XanderVertegaal:
This PR enables the user to add their own problems and save them to the DB. Problems added by the user can be edited by going into a special 'Edit' mode. Edits are now made to the Problem objects directly; in the future we might use annotations instead.

KB items are now also editable/viewable. They were not properly queried before.


backend/problem/views/problem.py
```diff
@@ -63,12 +74,194 @@ def get(self, request: Request, problem_id: int | None = None):
...
+
+
+class ParseInput(TypedDict):
+    id: str
+    premises: list[str]
+    hypothesis: str
+    kbItems: list[KBItem]
+
+
+def validate_input(parse_input: dict) -> ParseInput:
```
XanderVertegaal: Maybe this is a bit overkill or too verbose. Let me know if you have a better suggestion :)

backend/problem/models.py
```diff
@@ -73,13 +73,16 @@ def serialize(self) -> dict:
             case _:
                 serialized_extra_data = {}
 
+        kb_items = KnowledgeBase.objects.filter(problem=self)
```
bbonf: `self.knowledge_bases.all()` is slightly nicer

backend/problem/views/problem.py
```diff
@@ -63,12 +75,202 @@ def get(self, request: Request, problem_id: int | None = None):
...
             problem=problem,
+            index=problem_index,
             first=related_problem_ids.first,
             previous=related_problem_ids.previous,
             next=related_problem_ids.next,
             last=related_problem_ids.last,
             total=related_problem_ids.total,
         ).json_response(status=200)
+
+    def post(self, request, problem_id: int | None = None) -> JsonResponse:
```
bbonf: This code is fine, but you could consider refactoring it into something like:

```
def save_problem():
    ...

def post():
    try:
         self.save_problem()
    except Exception as e:
         return SaveProblemResponse(error=...)
```

where exceptions are allowed to be thrown from `save_problem()`

backend/problem/views/problem.py
bbonf: Your approach is not bad per-se, but its disadvantage is that the validation code is decoupled from the model. This means it's important to remember to update the validation code whenever the model updates. Maybe look into DRF's validators and serializers which can do this for you: https://www.django-rest-framework.org/api-guide/validators/

backend/problem/views/problem.py
```diff
@@ -63,12 +75,202 @@ def get(self, request: Request, problem_id: int | None = None):
...
+        hypothesis_sentence = Sentence.objects.get_or_create(
+            text=parse_input["hypothesis"]
+        )[0]
+    except Exception as e:
+        raise ValueError(f"Error creating hypothesis sentence: {e}")
+
+    problem = Problem.objects.create(
+        hypothesis=hypothesis_sentence,
+        dataset=Problem.Dataset.USER,
+        # TODO: Determine entailment label based on LangPro parser output.
```
bbonf: I think we want the user to supply the entailment label that they expect

backend/problem/views/problem.py
```diff
@@ -63,12 +75,202 @@ def get(self, request: Request, problem_id: int | None = None):
...
+    update_or_create_kb_items(problem=problem, kb_items=parse_input["kbItems"])
+
+    return problem
+
+
+def update_or_create_kb_items(problem: Problem, kb_items: list[KBItem]) -> None:
+    kb_ids: list[str] = []
+    for item in kb_items:
+        id = item.get("id")
+        entity1 = item.get("entity1")
```
bbonf: I'm not sure why this isn't `item["entity1"]` etc

backend/problem/views/problem.py
```diff
@@ -63,12 +75,202 @@ def get(self, request: Request, problem_id: int | None = None):
...
+            try:
+                kb = KnowledgeBase.objects.create(
+                    entity1=entity1,
+                    relationship=relationship,
+                    entity2=entity2,
+                    problem=problem,
+                )
+                kb_ids.append(kb.pk)
+            except Exception as e:
+                raise ValueError(f"Error creating knowledge base items: {e}.")
```
bbonf: This is the kind of exceptions that IMO can be left unhandled. If django fails so badly that it can't create an object, might as well crash hard. But your choice to handle it is also perfectly valid.

backend/problem/views/problem.py
```diff
@@ -63,12 +75,202 @@ def get(self, request: Request, problem_id: int | None = None):
...
+            kb_ids.append(kb.pk)
+
+    # Delete existing knowledge bases associated to this problem that are
+    # not included in the input.
+    KnowledgeBase.objects.filter(problem_id=problem.pk).exclude(id__in=kb_ids).delete()
+
+
+def update_problem_from_input(parse_input: ParseInput) -> Problem:
+    try:
+        problem = Problem.objects.get(id=parse_input["id"])
```
bbonf: This query should also filter on `USER` problems for extra safety (otherwise you can modify the request parameter and potentially overwrite any problem)

backend/problem/views/problem.py
```diff
@@ -63,12 +75,202 @@ def get(self, request: Request, problem_id: int | None = None):
...
+    except Problem.DoesNotExist:
+        raise ValueError(f"Cannot find Problem with ID: {parse_input["id"]}")
+
+    problem.hypothesis = Sentence.objects.get_or_create(text=parse_input["hypothesis"])[
+        0
```
bbonf: interesting indentation choice :)

frontend/src/app/annotate/annotation-input/knowledge-base-form/knowledge-base-form.component.html
```diff
@@ -19,12 +21,14 @@
                     focusable="false"
                 ></fa-icon>
             </button>
+            }
         </legend>
     </fieldset>
 
     <ul class="ps-0 list-unstyled" formArrayName="kbItems">
-        @for (kbItem of kbControls; let i = $index; track $index) {
+        @for (kbItem of kbControls; let i = $index; track `${i}-${kbItem.value.entity1}-${kbItem.value.relationship}-${kbItem.value.entity2}`) {
```
bbonf, line 29: This seems excessive but I admit I never fully understood how you're supposed to use `track`

frontend/src/app/services/app-mode.service.ts
```diff
@@ -0,0 +1,32 @@
...
+export enum AppMode {
+    BROWSE = "browse",
+    EDIT = "edit",
+    ADD = "add",
+}
+
+@Injectable({
+    providedIn: 'root'
+})
+export class AppModeService {
```
bbonf: I don't mean it in a harsh way but I don't really understand why this has to be a service. Is it not a state of the annotation component?
If it has to be a service, can't it be part of the problem service?

backend/problem/models.py
XanderVertegaal: I know... I used this syntax to keep the type checker happy, but I agree it's not worth it. I'll use your suggestion!

backend/problem/views/problem.py
XanderVertegaal: I see what you mean. I'll look into serializers for this. It looks promising!

backend/problem/views/problem.py
XanderVertegaal: Black makes these choices for me 😂

backend/problem/views/problem.py
XanderVertegaal: Ah great, if I let all exceptions bubble up, I can catch them all with one `try...except` block. It's much cleaner this way :)

backend/problem/views/problem.py
XanderVertegaal: I never considered crashing 'hard' (😂) to be a good thing, but you're right that it would be good to have a dramatic crash if something is up. I was thinking of handling cases in which KBItems cannot be created due to validation rules (e.g. uniqueness constraints), but that is not relevant for KBItems.

backend/problem/views/problem.py
XanderVertegaal: I thought that would belong in the realm of annotations (i.e. 'overriding' LangPro output) rather than editing a problem directly, but I see what you mean. It is not really clear to me where annotation ends and directly editing starts.

frontend/src/app/annotate/annotation-input/knowledge-base-form/knowledge-base-form.component.html
XanderVertegaal: You're supposed to give a unique simple value here. For static lists, `i` would work, but this fails if you delete the middle of three items, for instance. (The last one disappears from the UI.) For items with a unique ID, you can use the ID, but our list may contain multiple items whose ID is "new".

The only way I've found to create a unique ID for each item is to use both the index and all relevant values. I agree it's not pretty... :(

frontend/src/app/services/app-mode.service.ts
XanderVertegaal: It definitely could, and it was until I put it here :) I saw this more as global application state and wanted to keep it apart from other logic. I do agree that there is not a lot of use case for it outside of the AnnotationComponent. I'll put it back where it was!

XanderVertegaal:
Alright, I think I've got everything! Let me know if you think this is ready :)

backend/problem/problem_details.py
```diff
@@ -10,16 +10,16 @@
 
 @dataclass
 class RelatedProblemIds:
-    first: Optional[str] = None
-    previous: Optional[str] = None
-    next: Optional[str] = None
-    last: Optional[str] = None
-    total: Optional[int] = None
+    first: int | None = None
```
jgonggrijp, line 12: Scanning through the commits, I think I understand you replaced `str` by `int` for consistency. But, to complete my contextual understanding: why use `int | None` rather than `Optional[int]`?

backend/problem/problem_details.py
```diff
@@ -70,7 +70,7 @@ def get_filters(query_params: QueryDict) -> Q | None:
         filters &= Q(dataset=dataset)
     if entailment_label:
         filters &= Q(entailment_label=entailment_label)
-    if gold is not None:
+    if gold is not None and gold != "":
```
jgonggrijp: Isn't that just

```suggestion
    if gold:
```

backend/problem/tests.py
jgonggrijp, line 1: Ideally write these tests using [pytest](https://docs.pytest.org/en/stable/) notation instead. This is the convention in all our cookiecutter applications and is also used in this project already, for example in `/backend/user/tests/`. The gist is this: free-standing test functions instead of methods of `TestCase` subclasses and free-standing reusable fixtures instead of `setUp`/`tearDown` class methods. Fixtures are injected into test functions by naming them as test function parameters. Assertions are written using Python's built-in `assert` keyword and when they fail, the entire expression is pretty-printed to stdout.

jgonggrijp: If you convert this to pytest, the module should probably be renamed to `serializers_test.py`, because you seem to be specifically testing the behavior of the serializers you wrote. The cookiecutter convention puts backend modules and their tests side by side, just like in the frontend.

jgonggrijp, line 1: Good to have comprehensive tests for all this validation, by the way!

backend/problem/views/problem.py
```diff
@@ -1,30 +1,30 @@
...
-    first: Optional[str] = None
-    previous: Optional[str] = None
-    next: Optional[str] = None
-    last: Optional[str] = None
-    total: Optional[int] = None
+    first: int | None = None
+    previous: int | None = None
+    next: int | None = None
+    last: int | None = None
+    total: int | None = None
```
jgonggrijp: These are the same as in the problem model. Maybe factor them out as a shared mixin?

backend/problem/views/problem.py
```diff
@@ -38,8 +38,21 @@ def json_response(self, status=200) -> JsonResponse:
...
+    id: int | None = None
+    error: str | None = None
+
+    def json_response(self, status=200) -> JsonResponse:
+        return JsonResponse(asdict(self), status=status)
+
+
 class ProblemView(APIView):
-    def get(self, request: Request, problem_id: int | None = None):
+    def get(self, request, problem_id: int | None = None) -> JsonResponse:
```
jgonggrijp: Why remove the type annotation for `request` here?

backend/problem/views/problem.py
```diff
@@ -63,12 +76,146 @@ def get(self, request: Request, problem_id: int | None = None):
...
             previous=related_problem_ids.previous,
             next=related_problem_ids.next,
             last=related_problem_ids.last,
             total=related_problem_ids.total,
         ).json_response(status=200)
+
+    def post(self, request, problem_id: int | None = None) -> JsonResponse:
+        """
+        If the Problem ID is None, attempts to create a new Problem;
+        else the associated Problem is updated.
```
jgonggrijp: Is updating not usually done with `put` or `patch`?

backend/problem/views/problem.py
```diff
@@ -63,12 +76,146 @@ def get(self, request: Request, problem_id: int | None = None):
...
+        except DatabaseError as dbe:
+            logger.error(f"Database error saving problem: {dbe}")
+            return SaveProblemResponse(
+                id=problem_id, error="Database error saving problem."
+            ).json_response(status=500)
+        except Exception as e:
+            logger.exception(f"Unexpected error saving problem: {e}")
+            return SaveProblemResponse(
+                id=problem_id, error="Error saving problem."
+            ).json_response(status=500)
```
jgonggrijp: DRF already handles a lot of exceptions by automatically returning an appropriate HTTP response. I am not sure that all of this code is redundant, but you might want to read up on it.

https://www.django-rest-framework.org/api-guide/exceptions/

backend/problem/views/problem.py
```diff
@@ -63,12 +76,146 @@ def get(self, request: Request, problem_id: int | None = None):
...
+
+        return SaveProblemResponse(id=problem.pk).json_response(status=200)
+
+
+def save_problem(input_data: dict, problem_id: int | None) -> Problem:
+    serializer = ProblemInputSerializer(data=input_data)
+
+    if not serializer.is_valid():
+        raise ValueError("Input data is not valid.")
```
jgonggrijp: With DRF, you normally shouldn't need to manually invoke the serializer's validation. You can just declare the `serializer_class` of a view and it should handle almost everything automatically. But maybe you need to upgrade to a [`GenericAPIView`](https://www.django-rest-framework.org/api-guide/generic-views/) or a [`ModelViewSet`](https://www.django-rest-framework.org/api-guide/viewsets/) in order to get all the benefits.

backend/problem/views/problem.py
```diff
@@ -63,12 +76,146 @@ def get(self, request: Request, problem_id: int | None = None):
...
+                problem=problem,
+            )
+            kb_ids.append(kb.pk)
+        else:
+            kb = KnowledgeBase.objects.get(id=id, problem_id=problem.pk)
+            kb.entity1 = entity1
+            kb.relationship = relationship
+            kb.entity2 = entity2
+            kb.save()
+            kb_ids.append(kb.pk)
```
jgonggrijp: This code looks like something that should be handled in the problem serializer. [Yep](https://www.django-rest-framework.org/api-guide/relations/#writable-nested-serializers).

frontend/src/app/annotate/annotate.component.html
```diff
@@ -1,22 +1,93 @@
+@let firstProblemId = firstProblemId$ | async;
+@let appMode = appMode$ | async;
+@let browsing = appMode === 'browse';
+@let adding = appMode === 'add';
+@let editing = appMode === 'edit';
```
jgonggrijp, line 5: Do these dynamically change along with `appMode`? Nice, I knew you could do this in nested elements but I didn't know yet that this also worked with global `@let`.

frontend/src/app/annotate/annotate.component.ts
```diff
@@ -15,10 +21,51 @@ import { faTree } from "@fortawesome/free-solid-svg-icons";
...
+    public appMode$ = this.problemService.appMode$;
+    public firstProblemId$ = this.problemService.firstProblemId$;
+
+    public isUserProblem$ = this.problemService.problem$.pipe(
+        map(problem => problem?.dataset === Dataset.USER)
+    );
+
+    ngOnInit(): void {
+        const editParam$ = this.route.url.pipe(
+            map(segments => segments.some(segment => segment.path === "edit")),
```
jgonggrijp, line 47: I don't actually recommend that you do this because the project is not yet using Underscore, but I always like to point out nice ways to enable shorter notation like this:

```suggestion
            map(segments => _.some(segments, {path: "edit"})),
```

(You could even do `map(_.partial(_.some, _, {path: 'edit'}))`, though that is admittedly less recognizable for most people.)

frontend/src/app/annotate/annotate.component.html
```diff
@@ -1,22 +1,93 @@
...
                 ></fa-icon>
-                <span class="ms-4 fs-3" i18n>Start parse</span>
+                <span class="ms-3" i18n>Add new problem</span>
             </button>
+            } @else {
+            <button
+                type="button"
+                class="btn btn-secondary d-flex align-items-center justify-content-center"
+                (click)="firstProblemId && goToProblem(firstProblemId)"
+            >
```
jgonggrijp: I'm a bit rusty on Angular, so I might be going against the conventions and/or asking for the impossible, but my acquired preferences would tell me to rather do something like this instead,

```suggestion
            @if (firstProblemId) {
            <a
                class="btn btn-secondary d-flex align-items-center justify-content-center"
                href="/annotate/{{firstProblemId}}"
            >
```

so I wouldn't need to inject the router in the component and basically do the router's work by making it navigate manually. Is that conceivable?

jgonggrijp: BTW this is an incomplete change suggestion because I was too lazy to go all the way to the closing `</a>}`.

frontend/src/app/annotate/annotation-input/annotation-input.component.spec.ts
```diff
@@ -33,4 +44,147 @@ describe("AnnotationInputComponent", () => {
...
+                        id: 456,
+                        entity1: "cat",
+                        entity2: "animal",
+                        relationship: KnowledgeBaseRelationship.SUBSET
+                    },
+                    {
+                        id: 789,
+                        entity1: "dog",
+                        entity2: "pet",
+                        relationship: KnowledgeBaseRelationship.EQUAL
```
jgonggrijp, line 66: All pets are dogs???

frontend/src/app/annotate/annotation-input/annotation-input.component.ts
```diff
@@ -130,7 +176,23 @@ export class AnnotationInputComponent implements OnInit {
...
+                nonNullable: true
+            }),
+            entity2: new FormControl<string>(item.entity2, {
+                nonNullable: true
+            }),
+            relationship: new FormControl<KnowledgeBaseRelationship>(item.relationship, {
+                nonNullable: true
+            })
+        }));
+    }
```
jgonggrijp, line 194: Why isn't this job done by the template?

frontend/src/app/annotate/annotation-input/premises-form/premises-form.component.html
```diff
@@ -1,11 +1,13 @@
 @let premiseControls = form().controls.premises.controls;
+@let appMode = appMode$ | async;
 
 <form [formGroup]="form()">
     <fieldset>
         <legend
             class="form-label h6 fw-bold d-flex align-items-center justify-content-between"
         >
             <span i18n>Premises</span>
+            @if (appMode === 'add' || appMode === 'edit') {
```
jgonggrijp: You use this expression a lot in this template. Maybe assign it to its own `@let`?

jgonggrijp (commented):
I have not read all parts of the diff with equal attention, because it is a large diff. I also do not really have an opinion on your overall approach, because I am still very new to the project and also because I am a bit rusty on Angular. That being said, I could recognize what you've done as basically adjusting and extending existing code to accomodate a new problem type. As such, the changes look reasonable to me. I also like that you added a substantial amount of nontrivial unittests.

Apart from that, I noticed that you don't seem very familiar yet with DRF and with pytest. You might like to try the [long-form DRF tutorial](https://www.django-rest-framework.org/tutorial/1-serialization/) and pytest's [quickstart](https://docs.pytest.org/en/stable/getting-started.html) and [how-to guides](https://docs.pytest.org/en/stable/how-to/index.html) (I especially recommend the one on fixtures).

backend/problem/problem_details.py
XanderVertegaal: I prefer the more modern `int | None` syntax (PEP 604) over `Optional[int]`. (It's more similar to TS unions and avoids an import from `typing`.) The new syntax only works on Python 3.10+, I believe, which is why we started out using `Optional[int]`.

backend/problem/views/problem.py
XanderVertegaal: You're right! I'll adjust it.

frontend/src/app/annotate/annotate.component.html
XanderVertegaal: Yes! It's a relatively new feature (stable since Angular 19, I believe). I think this is much cleaner than using a bunch of `*ngIf`s.

backend/problem/views/problem.py
XanderVertegaal: Great, TIL!

backend/problem/views/problem.py
XanderVertegaal: Done!

frontend/src/app/annotate/annotate.component.html
XanderVertegaal: No, you're right! All this button does is navigating, so I agree it is best replaced by an anchor tag.

XanderVertegaal: Same goes for the one above it. I'll run a global check to see if other buttons in this PR can be replaced with `<a>`s for semantic reasons.

frontend/src/app/annotate/annotate.component.ts
XanderVertegaal: Interesting. Would the type checker know about `path` has to be a property of an element of `segments`? Would it complain if I accidentally wrote `ptah` (or some other Egyptian god)?

frontend/src/app/annotate/annotation-input/annotation-input.component.spec.ts
XanderVertegaal: The tests pass, so it must be true!!! (Don't tell Luke.)

frontend/src/app/annotate/annotation-input/annotation-input.component.ts
XanderVertegaal: What do you mean? This method builds a set of `FormGroup`s based on data received from the backend. I don't see how the template would be able to accomplish this.

frontend/src/app/annotate/annotation-input/premises-form/premises-form.component.html
XanderVertegaal: Done!

frontend/src/app/annotate/annotate.component.ts
jgonggrijp: The people who wrote `@types/underscore` (and `/lodash`) went to great lengths to enable TS to infer such things. In this case, I think it might be able to catch the typo. That being said, JS is at its core a dynamically typed language and some of its types are just untraceable without running the program. Underscore has not restricted itself to the subset of JS that TS can accurately type. For example, I know that `_.flatten` has a type that TS cannot express because of the nested infinities involved.

frontend/src/app/annotate/annotation-input/annotation-input.component.ts
jgonggrijp: Aren't templates all about generating HTML dynamically?

XanderVertegaal: Well, on the basis of the `FormGroup`s created here, the template of`KnowledgeBaseFormComponent` adds input/select/button elements as required. I think this is a standard Angular pattern.

Let me know if you have another pattern in mind!

jgonggrijp: You know the Angular conventions much better than I do, so I'm probably missing something. Still, to answer your question: so far my impression is that we request a resource from a server, the response goes into an observable, usually with some transformations and intermediate steps, and eventually ends up being rendered in a template, for example through an `async` pipe. Along the way, components get created because their names appear somewhere in a template. We do a lot of administration so that the rendering engine that ships with Angular can figure out which components should be inserted where.

Over here, you seem to be detaching yourself from that practice and creating the components manually from the JavaScript side instead. It seems a bit like the Angular equivalent of calling `document.createElement('div')` instead of writing `<div>` in a HTML file or template. There are sometimes good reasons to do so, but it is not clear to me yet what those reasons are in this case.

XanderVertegaal: I see what you mean. In this case, we don't know yet how many KB forms we will need to display on the page, as the problem that is retrieved from the backend may contain zero, one or multiple KB items, so I can't hardcode the HTML.

I could map the individual KB items from the backend straight to HTML input elements, but then I can't easily interact with them or validate them. That is why I am mapping them to FormGroups, which hold not only values but also validation rules and provide utilities to interact with your form (mark all controls as touched etc.). This is all in accordance with the [Reactive Forms](https://angular.dev/guide/forms/reactive-forms) pattern, which is more flexible than the Template-driven pattern.

I guess you could say the FormGroup is the Model in the MVC pattern and the HTML inputs are the View that is generated dynamically on the basis of the Model, but I'm not well acquainted with the MVC pattern yet to say how accurate that statement is 😅

jgonggrijp: In a template you can handle "don't know how many yet" using an `@for` loop, no?

FormGroup is probably more like a viewmodel. Angular components are often viewmodels, which is why they've declared themselves to be an MVVM framework. (MVC and MVVM are basically the same paradigm, just with different emphasis in the naming.)

I will read up on Reactive Forms. Let's consider that the resolution of this thread.

PR #57, opened 2025-09-22
Add Icon Button Component

XanderVertegaal:
No functional changes. I have only replaced button + icon combinations with a custom component.

### DIFF
--- a/frontend/src/app/shared/icon-button/icon-button.component.html
+++ b/frontend/src/app/shared/icon-button/icon-button.component.html
@@ -1,0 +1,29 @@
+@let iconValue = icon(); @let position = iconPosition();
+<button
+    [type]="buttonType()"
+    [class]="buttonClasses()"
+    [disabled]="disabled()"
+    [attr.aria-labelledby]="titleId() ?? ''"
+    [attr.title]="buttonTitle()"
+>
+    @if (iconValue && position === 'left') {
+    <fa-icon
+        [icon]="iconValue"
+        [class]="`fa-${size()} ${label() ? 'me-2' : ''}`"
+        aria-hidden="true"
+        focusable="false"
+    />
+    }
+    <span i18n>{{ label() }}</span>
+    @if (iconValue && position === 'right') {
+    <fa-icon
+        [icon]="iconValue"
+        [class]="`fa-${size()} ${label() ? 'ms-2' : ''}`"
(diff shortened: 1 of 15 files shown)


frontend/src/app/shared/icon-button/icon-button.component.html
```diff
@@ -0,0 +1,31 @@
...
+    [type]="buttonType()"
+    [class]="buttonClasses()"
+    [disabled]="disabled()"
+    [attr.aria-labelledby]="titleId()"
+    [attr.title]="buttonTitle()"
+>
+    @if (iconValue && position === 'left') {
+    <fa-icon
+        [icon]="iconValue"
+        [class]="`fa-${size()} ${label() ? 'me-2' : ''}`"
```
jgonggrijp, line 12: "me too!"

frontend/src/app/shared/icon-button/icon-button.component.ts
```diff
@@ -0,0 +1,52 @@
...
+    public outline = input(false, { transform: booleanAttribute });
+    public fullWidth = input(false, { transform: booleanAttribute });
+
+    public buttonClasses = computed<string>(() => {
+        const classes = ["btn", "d-flex", "align-items-center", "justify-content-center"];
+        if (this.outline()) {
+            classes.push(`btn-outline-${this.buttonStyle()}`);
+        } else {
+            classes.push(`btn-${this.buttonStyle()}`);
+        }
```
jgonggrijp, line 29:
```suggestion
        const classInsert = this.outline() ? '-outline' : '';
        classes.push(`btn${classInsert}-${this.buttonStyle()}`);
```

jgonggrijp (commented):
I feel a bit conflicted about this change. In general, I'm all in favour of outfactoring repeated patterns, and I do appreciate that you are able to simplify the templates because of the new component. On the other hand, this approach solidifies the pattern that I objected against in #53, where you use a `<button (click)="action()">` and a manual `router.navigate()` for what is essentially a `<a href="/target/route">`.

Hence, no verdict; I think you are better able to judge whether this should be merged. We can also discuss this topic if you think that might be useful.

XanderVertegaal:
I see your point and I agree that this may not be a good pattern to adopt. I'll close this PR now; maybe we can simplify these links/buttons some other way in the future.

Issue #12, opened 2025-03-18
User permissions


XanderVertegaal:
There are going to be two types of users:
- Annotators: should be allowed to annotate
- Master annotators (vel sim.): should be allowed to annotate and mark annotations/problems as gold.

XanderVertegaal:
Update: 3 kinds of users (in addition to superusers/admins):

- Registered users can
  - browse problems and parse results but not edit them.

- Annotators can
  - add/edit KB items (of anyone) for non-locked problems
  - apply categories/labels to non-locked problems
  - remove their own categories/labels to non-locked problems
  - browse locked problems

- Master Annotators can
  - lock problems / mark them as gold
  - CRUD categories/labels
  - add users to annotator user group

XanderVertegaal:
Update (14 Nov 2025):

|                            | Visitor | Annotator | Master Annotator |
| -------------------------- | ------- | --------- | ---------------- |
| Browse gold problems       | Yes     | Yes       | Yes              |
| Browse silver problems     | No      | Yes       | Yes              |
| Edit KB items              | No      | Yes       | Yes              |
| Add labels                 | No      | Yes       | Yes              |
| Remove own labels          | No      | Yes       | Yes              |
| Remove other users' labels | No      | No        | Yes              |
| Copy problems              | No      | No        | Yes              |
| See hidden problems        | No      | No        | Yes              |
| Silver/gold problems       | No      | No        | Yes              |
| Hide/unhide problems       | No      | No        | Yes              |
| Manage users               | No      | No        | Yes              |

XanderVertegaal:
Update 12 Dec 2025

|                        | Unregistered users | Registered Users | Annotators | Master Annotators |
| ---------------------- | ------------------ | ---------------- | ---------- | ----------------- |
| Browse gold problems   | Yes                | Yes              | Yes        | Yes               |
| Browse silver problems | No                 | Yes              | Yes        | Yes               |
| Browse bronze problems | No                 | Yes              | Yes        | Yes               |
| Edit KB items          | No                 | No               | Yes        | Yes               |
| Add labels             | No                 | No               | Yes        | Yes               |
| Remove own labels      | No                 | No               | Yes        | Yes               |
| Remove others' labels  | No                 | No               | No         | Yes               |
| Copy problems          | No                 | No               | No         | Yes               |
| See hidden problems    | No                 | No               | No         | Yes               |
[...]

PR #61, opened 2025-11-04
Feature/user models

XanderVertegaal:
Addresses part of #12 

Adds a `role` field to the User model. I opted to add a field `canEditOrAddProblem` to the serializer so we don't have to implement a method that checks whether a user with role X can do Y in two places.


backend/user/models.py
```diff
@@ -1,14 +1,33 @@
...
+    # a UserProfile model.
+
+    class Meta:
+        db_table = "auth_user"
+
+    class Role(models.TextChoices):
+        ANNOTATOR = "annotator", "Annotator"
+        MASTER_ANNOTATOR = "master_annotator", "Master Annotator"
+        VISITOR = "visitor", "Visitor"
+
```
jgonggrijp: Aren't you basically reinventing and hardcoding user groups here?

jgonggrijp (commented):
#12 has a table of permissions that each role should or should not have. Why not just define those as custom permissions (where needed) and then implement the roles as regular user groups? That way you don't have to hardcode the roles and Django will assist more in enforcing the permissions.

XanderVertegaal:
Update! At your suggestion, the role field has been replaced by Django user groups and permissions. I have added a data migration that creates them and assigns the required permissions to them. For now there are only two groups: 'annotators' and 'master annotators'. Technically there is a third group (now called 'visitors' in the frontend) for users who are logged in but have not been assigned to either user group. However, these do not (yet?) have any special permissions that non-authenticated users lack, so I did not create a separate group for them now.

jgonggrijp:
Would you like me to take another look?

XanderVertegaal:
If you have time, yes please!

backend/problem/views/problem_test.py
```diff
@@ -0,0 +1,293 @@
...
+
+from problem.models import Problem, Sentence
+from user.models import User, GroupName
+from user.permissions import ANNOTATOR_PERMISSIONS, MASTER_ANNOTATOR_PERMISSIONS
+
+
+@pytest.fixture
+def api_client():
+    """Returns a DRF APIClient instance."""
+    return APIClient()
```
jgonggrijp: You don't need to create your own client fixture. `pytest-django`, which is already installed because of the cookiecutter, ships with a `client` fixture out of the box.

backend/problem/views/problem_test.py
```diff
@@ -0,0 +1,293 @@
...
+    ):
+        """Master annotators should be able to update user-created problems."""
+        api_client.force_authenticate(user=master_annotator)
+        response = api_client.patch(
+            f"/api/problem/{sample_problem.id}/", problem_input_data, format="json"
+        )
+        assert response.status_code == status.HTTP_200_OK
+
+
+class TestUserRoleProperties:
```
jgonggrijp, line 249: Terrible nitpick, I know, but Python calls them "attributes".

backend/problem/models.py
```diff
@@ -56,8 +75,6 @@ def get_index(self, qs: QuerySet) -> int | None:
         except Exception as e:
             logger.exception(f"Error getting index for problem {self.pk}: {e}")
             return None
-
-
```
jgonggrijp, line 60: Did you actually remove two blank lines here, or is GitHub's diff viewer fooling me?

backend/user/admin.py
```diff
@@ -1,8 +1,11 @@
-from django.contrib import admin
-from django.contrib.auth import admin as auth_admin
-from . import models
-
-
-@admin.register(models.User)
-class UserAdmin(auth_admin.UserAdmin):
-    pass
+from typing import Any
```
jgonggrijp: This import seems unused.

backend/user/admin.py
```diff
@@ -1,8 +1,11 @@
...
+from django.contrib import admin
+from django.contrib.auth import admin as auth_admin
+from django.http import HttpRequest
```
jgonggrijp: and this one

backend/user/admin.py
```diff
@@ -1,8 +1,11 @@
...
+from . import models
+from django.utils.translation import gettext_lazy as _
```
jgonggrijp, line 4: and this one as well.

frontend/src/app/annotate/annotation-input/annotation-input.component.ts
```diff
@@ -41,15 +44,16 @@ export type ParseInput = ReturnType<ParseInputForm["getRawValue"]>;
...
+    CommonModule,
+    PremisesFormComponent,
+    KnowledgeBaseFormComponent,
+    FormsModule,
+    ReactiveFormsModule,
+    ProblemDetailsComponent,
+    FontAwesomeModule,
+    RouterLinkWithHref,
+    IconButtonComponent
+],
```
jgonggrijp: Seem to have lost some indentation here.

frontend/src/app/guards/can-edit-or-add.guard.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    // No need to check permissions if just viewing.
+    if (!(editing || adding)) {
+        return true;
+    }
+
+    return authService.currentUser$.pipe(
+        // Wait until we actually have a user (User | null).
+        filter(user => user !== undefined),
+        take(1),
```
jgonggrijp, line 20: Isn't this a security risk? What if an annotator signs in, then logs out, and in the same tab a ~plebeian~visitor signs in?

frontend/src/app/menu/user-menu/user-menu.component.ts
```diff
@@ -1,72 +1,89 @@
...
+                case UserRole.SUPERUSER:
+                    return faUserAstronaut;
+                case UserRole.ANNOTATOR:
+                    return faUserTag;
+                case UserRole.MASTER_ANNOTATOR:
+                    return faUserGraduate;
+                case UserRole.VISITOR:
+                    return faUser;
+            }
+        })
```
jgonggrijp: If you write this at module scope,

```js
const defaultUserIcon = faUser;
const roleIcons = {
    [UserRole.SUPERUSER]: faUserAstronaut,
    [UserRole.ANNOTATOR]: faUserTag,
    [UserRole.MASTER_ANNOTATOR]: faUserGraduate,
    [UserRole.VISITOR]: defaultUserIcon,
};

const user2icon = user => user && roleIcons[user.role] || defaultUserIcon;
```

you can write this here:

```suggestion
        map(user2icon)
```

frontend/src/app/services/problem.service.ts
```diff
@@ -74,17 +78,44 @@ export class ProblemService {
...
+                        })) ?? []
+                    }
+                };
+            }));
+        }
+        return of<ProblemResponse>({
+            ...sharedProblemResponse,
             problem: {
                 id: null,
+                base: baseParam,
```
jgonggrijp: I was going to comment that you should refactor this, but I vaguely recall commenting on very similar code before. Did you already change this on another branch?

jgonggrijp (changes requested):
I like the comprehensive tests!

I have some comments, but it's mostly questions or nitpicks.

backend/problem/views/problem_test.py
XanderVertegaal: On the Python level they may be attributes, but in Django they are marked as 'properties' (cf. the `@property` decorator used in the definition of these fields) so if you do not mind I will be sticking to `TestUserRoleProperties` in this case.

In general, by the way, I appreciate nitpicks like this one; please keep them coming!

frontend/src/app/menu/user-menu/user-menu.component.ts
XanderVertegaal: Done! <3

frontend/src/app/services/problem.service.ts
XanderVertegaal: If I recall correctly, it was in a comment that said something along the lines of: "I don't recommend you change this, but if we used 'underscore', you could do this: ...". Did you change your mind about refactoring?

frontend/src/app/guards/can-edit-or-add.guard.ts
XanderVertegaal: This guard is triggered every time the user navigates to a page on which it is active. After logging in as a ~plebeian~ visitor, you will have to navigate back to an add/edit page, which will trigger this guard and decide another whether to let the user go ahead and login through a new subscription.

The `take(1)` is simply here to make sure that the subscription is closed after receiving the first emission of `User` or `null`. It doesn't mean that the subscription stays alive after this guard has run.

backend/problem/models.py
XanderVertegaal: The diff viewer is right, but the two blank lines should be here! I've reinstated them.

frontend/src/app/services/problem.service.ts
jgonggrijp: hmm maybe refactor without underscore? Just to get rid of the deep nesting?

frontend/src/app/shared/user2icon.ts
jgonggrijp, line 1: I noticed that every time I write that you can outfactor something "at module scope", you implement this in a new utility module. As far as I'm concerned, you could also just keep it in the module where the logic originated. Maybe you're already aware of this and you just prefer a separate module, but I thought I'd mention it.

PR #62, opened 2025-11-13
Docker environment for development

jgonggrijp:
Since I might step in to contribute to this project in the near future, I went ahead and prepared the project for development using Docker. This enables me to work on the project without having to install all the dependencies on my development machine.

These changes are NOT intended for deployment, only for development.

I have not yet looked into running the functional tests in Docker. This is probably possible, but seemed less urgent. However, I have verified that I can run the unittests of both frontend and backend.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ venv.bak/
 # Node
 node_modules/
 frontend/src/environments/version.ts
+*~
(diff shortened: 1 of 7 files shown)


bbonf (approved):
Nice, looks good to me (but I haven't tried running it)

PR #64, opened 2025-12-17
Feature/problem labels

### LINKED ISSUE #60: Problem categories/labels
Master annotators should be able to create new labels (= categories) in Django Admin, e.g. 'Wrong entailment label', 'Require change of sentence', 'Typo' etc.

Master annotators and annotators should be able to pick these labels and apply one or multiple of them to individual problems and add a comment justifying/explaining why they think it's appropriate.

Annotator users should only be able to remove their own labels, not those of other annotators. Master annotators should be able to remove any label.

Annotators/master annotators should be able to quickly filter problems based on (a combination of) labels.

XanderVertegaal:
Closes #60 

This is one key aspect of annotations accounted for. We may also be able to reuse this system to mark problems as 'gold' or 'hidden', although I still think these properties are best kept independent.

The fixtures in `backend/annotation/fixtures/initial.json` are for testing convenience only, so you don't have to add labels yourself every time you flush your local dev DB. They are not intended to be loaded in on the server.


backend/annotation/views.py
```diff
@@ -72,55 +72,70 @@ def create(self, request: Request) -> Response:
 
         selected_label_ids = {label["id"] for label in selected_labels}
 
+        try:
+            self._update_labelings(problem, user, selected_label_ids, remarks)
+        except PermissionError as e:
+            return Response({"detail": str(e)}, status=403)
```
jgonggrijp: I think error-handling behavior is already built into DRF.

backend/annotation/views.py
```diff
@@ -72,55 +72,70 @@ def create(self, request: Request) -> Response:
 
         selected_label_ids = {label["id"] for label in selected_labels}
 
+        try:
+            self._update_labelings(problem, user, selected_label_ids, remarks)
+        except PermissionError as e:
+            return Response({"detail": str(e)}, status=403)
+
+        return Response({"ok": True}, status=HTTP_200_OK)
```
jgonggrijp: Could just do

```suggestion
        return {"ok": True}
```

backend/annotation/models.py
```diff
@@ -81,3 +81,83 @@ class KnowledgeBaseAnnotation(models.Model):
...
+        help_text="When this label was removed from the problem.",
+    )
+    removed_by = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE,
+        related_name="labelings_removed",
+        null=True,
+        blank=True,
+        help_text="User who removed this label.",
+    )
```
jgonggrijp, line 138: Not a request for change, just a remark: another way to organize this would be to create a separate `Event` model for changes made by users. It would have a type (such as "deleted"), a link to the affected model, a user and a date/time. Maybe not worth worrying about for this project, but when there many models that can be edited, this could help to keep the models leaner.

backend/annotation/views_test.py
```diff
@@ -0,0 +1,290 @@
...
+        assert response.status_code == status.HTTP_200_OK
+
+
+class TestSaveLabelingsPermissions:
+    """Tests for POST permissions (saving labelings) on the Label endpoint."""
+
+    def test_unauthenticated_user_cannot_save_labelings(
+        self, api_client, sample_problem, sample_label
+    ):
+        """Unauthenticated users should not be able to save labelings."""
```
jgonggrijp, line 91: I like that you consistently add docstrings to your test cases. I've been too lazy to do it with pytest because it didn't force me, but I think I'll try to adopt this habit from you.

frontend/src/app/annotate/annotation-input/problem-details/problem-labels/manage-labels-modal/manage-labels-modal.component.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+            tick();
+
+            const countAfterAdd = availableCount;
+
+            component.removeLabel(testAvailableLabel.id);
+            tick();
+
+            expect(availableCount).toBe(countAfterAdd + 1);
+        }));
+    });
```
jgonggrijp, line 218: These are tests for a component, which is a view and a controller in the MVC triangle, but I think the tested behavior belongs in the model layer. Indeed, it looks like something that could fairly easily exist indepently of the component.

This is **not** a change request&mdash;it's probably not important enough to hold up this PR&mdash;but it might be worth creating a ticket for, as something that we might (optionally) refactor later.

frontend/src/app/annotate/annotation-input/problem-details/problem-labels/manage-labels-modal/manage-labels-modal.component.ts
```diff
@@ -0,0 +1,136 @@
...
+            }
+            const selectedIds = selectedLabels.map(label => label.id);
+            return allLabels.filter(label => !selectedIds.includes(label.id));
+        })
+    );
+
+    public loadingLabels$ = this.availableLabels$.pipe(
+        map(() => false),
+        startWith(true)
+    );
```
jgonggrijp, line 66: This looks a bit hacky, though I don't have a ready suggestion for a different approach. Which other options did you consider, if any?

frontend/src/app/shared/formatTooltipText.ts
jgonggrijp, line 1: Why didn't you just include this in `../util.ts`?

frontend/src/app/shared/sortLabels.ts
jgonggrijp, line 1: Why didn't you just include this in `../util.ts`?

jgonggrijp (commented):
I have read the code but did not try out the features locally.

I like the comprehensive tests and documentation and the clear evidence that you aimed to work with standalone functions where possible. I have a few comments, but it all looks very reasonable to me.

backend/annotation/models.py
XanderVertegaal: Sounds very scalable! 👍

backend/annotation/views.py
XanderVertegaal: This works at runtime, but Pylance complains that this function now incorrectly overrides `CreateModelMixin.create`

```
Method "create" overrides class "CreateModelMixin" in an incompatible manner
  Return type mismatch: base method returns type "Response", override returns type "dict[str, bool]"
    "dict[str, bool]" is not assignable to "Response"
```

XanderVertegaal: As a compromise, I'll do `return Response({"ok": True})`

frontend/src/app/annotate/annotation-input/problem-details/problem-labels/manage-labels-modal/manage-labels-modal.component.spec.ts
XanderVertegaal: #68 has been added!

frontend/src/app/annotate/annotation-input/problem-details/problem-labels/manage-labels-modal/manage-labels-modal.component.ts
XanderVertegaal: The easiest solution here is to use the [`tap()` operator](https://rxjs.dev/api/operators/tap) and set a loading state variable to `false` once a request comes in. This is the imperative approach, relying on an unnamed side effect to update loading state. With more complex code it becomes difficult to keep track of what changes this `loading` boolean. In general I try to avoid `tap()` outside of debug logging.

Another solution is to wrap both the data and the loading state in a new object and to emit _two_ instances of that object: one with `{ loading: true, data: null }` followed by one with `{ loading: false, data: ... }`. This is more declarative but more verbose and complicated. We have adopted this solution in [the Lettercraft SearchService](https://github.com/CentreForDigitalHumanities/lettercraft/blob/0.12.0/frontend/src/app/services/search.service.ts)

The version in this file strikes a good position in the middle: it is declarative/reactive (which I prefer) yet does not suffer from the complexity of a `switchMap` into a `concat(of(...), ...)` as seen in Lettercraft.

backend/annotation/views.py
jgonggrijp: pylance is wrong.

frontend/src/app/annotate/annotation-input/problem-details/problem-labels/manage-labels-modal/manage-labels-modal.component.ts
jgonggrijp: Other question: does the state of label loading need to be a separate observable? Couldn't you just use `@let loadingLabels = (selectedLabels$ | async == null)`?

XanderVertegaal: You mean `availableLabels$`? With the new `@let` syntax I guess you could. I'm not sure whether you'd need to add a `startWith(null)` so the variable actually exists and has a null value before the first (real) data emission. (My database/branches are a bit messy right now so testing it would take quite some time.)

I don't have a real preference for either solution.

jgonggrijp: Er yes, `availableLabels$`. `undefined == null` is `true`, so I don't think you would need to `startWith(null)`. The only way this could go wrong, is if the initializer of `availableLabels$` only runs after the component is first rendered, but that is probably not the case since otherwise you also couldn't access `loadingLabels$`.

PR #63, opened 2025-11-14
Feature/save kb

bbonf:
@XanderVertegaal the work in this branch is not completed, but it might be useful if you can give this a look and see if this is still useful. or maybe you already worked on some overlapping functionality elsewhere.

PR #66, opened 2025-12-23
Feature/save kb reconcile

XanderVertegaal:
This PR builds on #63 and includes the latest version of develop. It develops the original PR's idea of using KnowledgeBaseAnnotations, bound by a AnnotationSession, instead of a KnowledgeBase model that is appended to a Problem directly. I think this is a good change which fits the researcher's vision that user annotations will mainly consist of additions/edits of knowledge base items and proof trees.

This PR is still a work in progress. Users can edit and save knowledge base items, but we need to decide where directly 'editing' a problem stops and where 'annotating' begins. Knowledge base edits currently take the form of annotations, but changes to sentences (premises + hypothesis) are directly made on the problem. Maybe we need both, or maybe we consider all changes to problems to be annotations.

In any case, this is something I'd like to clear up in discussion with you guys (@bbonf / @jgonggrijp)  before I finish this PR.


XanderVertegaal:
This PR has been redone in #71

PR #71, opened 2026-02-01
Feature/kb and labels as annotations

XanderVertegaal:
This PR builds on https://github.com/CentreForDigitalHumanities/langpro-annotator/pull/63 and includes the latest version of develop. It develops the original PR's idea of using KnowledgeBaseAnnotations, bound by a AnnotationSession, instead of a KnowledgeBase model that is appended to a Problem directly. I think this is a good change which fits the researcher's vision that user annotations will mainly consist of additions/edits of knowledge base items and proof trees.

AnnotationSessions also have a "notes" field aimed at remarks or justifications. I first implemented that too, but noticed that any notes given when adding a label are overwritten when a label is removed. This would have been easier to implement with the 'Event'/'Change' structure you proposed in an earlier PR review.


backend/annotation/models.py
```diff
@@ -96,68 +139,24 @@ def __str__(self):
         return self.text
 
 
-class Labeling(models.Model):
-    """
-    The attachment of a label to a problem.
-
-    Each time a label is attached to a problem, a new Labeling record is created.
-    When removed, the record is marked as removed (not deleted), so the history of labelings is preserved.
-    """
```
jgonggrijp, line 105: Could this docstring return elsewhere in an updated form?

backend/annotation/serializers.py
```diff
@@ -2,11 +2,99 @@
...
+        return user.has_perm("annotation.delete_knowledgebaseannotation")
+
+    def validate_id(self, value):
+        """Validate that the KnowledgeBaseAnnotation ID exists if provided."""
+        if value is not None:
+            if not KnowledgeBaseAnnotation.objects.filter(id=value).exists():
+                raise serializers.ValidationError(
+                    f"KnowledgeBaseAnnotation item with ID {value} does not exist."
+                )
+        return value
```
jgonggrijp: I've hesitated to comment on this because it seems a bit like hairsplitting, but I'm uncomfortable with the combination of nested and negated conditions. I would prefer this, which I think is much easier to digest:


```suggestion
        if value is None or KnowledgeBaseAnnotation.objects.filter(id=value).exists():
            return value
        raise serializers.ValidationError(
            f"KnowledgeBaseAnnotation item with ID {value} does not exist."
        )
```

backend/annotation/serializers.py
```diff
@@ -2,11 +2,99 @@
...
+
+    def update(
+        self, instance: KnowledgeBaseAnnotation, validated_data: dict
+    ) -> KnowledgeBaseAnnotation:
+        """Update an existing KnowledgeBaseAnnotation item."""
+        instance.entity1 = validated_data["entity1"]
+        instance.relationship = validated_data["relationship"]
+        instance.entity2 = validated_data["entity2"]
+        instance.save()
+        return instance
```
jgonggrijp: Isn't this just the default behavior of `ModelSerializer.update`?

backend/annotation/serializers.py
```diff
@@ -17,83 +105,89 @@ class Meta:
...
             return False
 
-        if user.is_superuser or user.has_perm("annotation.delete_any_labeling"):
+        if user.is_superuser or user.has_perm("annotation.delete_any_labelannotation"):
             return True
 
-        if user.has_perm("annotation.delete_own_labeling"):
-            return labeling.attached_by.pk == user.pk
+        if user.has_perm("annotation.delete_own_labelannotation"):
+            return annotation.created_by.pk == user.pk
```
jgonggrijp: This is a repetition of line 130. While it's only one line, the logic could change and in that case both lines would have to be updated, so maybe outfactor it to a shared method anyway?

backend/annotation/serializers.py
```diff
@@ -17,83 +105,89 @@ class Meta:
...
+
+    def _get_problem_and_last_session(self):
+        problem = self.context.get("problem", None)
+        user = self.context.get("user", None)
+        if not problem or not user or user.is_authenticated is False:
+            return None, None
+        last_session = (
+            AnnotationSession.objects.filter(user=user).order_by("-created_at").first()
+        )
+        return problem, last_session
```
jgonggrijp: This serializer confuses me. Why does it exist, what does it serialize the annotations of? Why are both the current problem and the most recent annotation session consulted? Why does this serializer seem to be concerned with constructing query sets, which would usually be a job for views and viewsets?

backend/annotation/serializers_test.py
```diff
@@ -0,0 +1,298 @@
...
+    assert "id" in serializer.errors
+
+
+@pytest.mark.django_db
+def test_kb_item_missing_relationship(kb_annotation):
+    """Test that a kbItem missing a relationship is invalid."""
+    data = {"id": kb_annotation.pk, "entity1": "e1", "entity2": "e2"}
+    serializer = KnowledgeBaseAnnotationSerializer(data=data)
+    assert not serializer.is_valid()
+    assert "relationship" in serializer.errors
```
jgonggrijp: You declaratively defined the relationship as required. You don't need to write your own tests for that kind of thing because you can rely on DRF already having tests in place to ensure that such a declaration is actually enforced.

Most of the other tests are certainly valuable, though!

backend/annotation/serializers_test.py
```diff
@@ -0,0 +1,298 @@
...
+        kb_annotation, context={"request": request}
+    )
+    data: dict[str, Any] = serializer.data  # type: ignore
+
+    assert data["id"] == kb_annotation.pk
+    assert data["entity1"] == "e1"
+    assert data["entity2"] == "e2"
+    assert data["relationship"] == "equal"
+    assert "createdAt" in data
+    assert "createdBy" in data
```
jgonggrijp: This is another case where you are basically testing whether DRF is working as advertised.

backend/annotation/serializers_test.py
```diff
@@ -0,0 +1,298 @@
...
+def test_label_annotation_removable_by_creator(label_annotation, annotator):
+    """Test that creator can remove their own annotation with proper permission."""
+    perm = Permission.objects.get(codename="delete_own_labelannotation")
+    annotator.user_permissions.add(perm)
+
+    serializer = LabelAnnotationSerializer(
+        label_annotation, context={"user": annotator}
+    )
+    data: dict[str, Any] = serializer.data  # type: ignore
+    assert data["removable"] is True
```
jgonggrijp, line 141: I'm missing a converse test, whether the creator can *not* remove their own label annotation *without* the permission.

backend/annotation/views_test.py
```diff
@@ -23,22 +23,36 @@ def another_label(db):
 
 
 @pytest.fixture
-def labeling_by_annotator(db, sample_problem, sample_label, annotator):
-    """Creates a labeling attached by an annotator."""
-    return Labeling.objects.create(
+def annotation_session_annotator(db, annotator):
+    """Creates an annotation session for the annotator."""
+    return AnnotationSession.objects.create(user=annotator)
```
jgonggrijp: Wouldn't that be an `annotator_session`?

backend/annotation/views_test.py
```diff
@@ -23,22 +23,36 @@ def another_label(db):
...
+
+
+@pytest.fixture
+def annotation_session_master(db, master_annotator):
+    """Creates an annotation session for the master annotator."""
+    return AnnotationSession.objects.create(user=master_annotator)
```
jgonggrijp: Likewise, a `master_annotator_session`?

backend/problem/serializers.py
```diff
@@ -130,11 +89,82 @@ def create(self, validated_data: dict) -> Problem:
...
         kb_items = validated_data.get("kbItems", [])
-        if kb_items:
-            self._update_or_create_kb_items(problem, kb_items)
+        labels = validated_data.get("labels", [])
+        if kb_items or labels:
+            self._handle_annotations(problem, kb_items, labels)
 
         return problem
 
+    def _handle_annotations(self, problem: Problem, kb_items: list[dict], labels: list[dict]) -> None:
```
jgonggrijp: This method is too complex, please refactor.

backend/problem/serializers.py
```diff
@@ -213,8 +221,17 @@ class ProblemInputSerializer(serializers.Serializer):
     hypothesis = serializers.CharField(
         allow_blank=False, help_text="Hypothesis sentence text"
     )
-    kbItems = KnowledgeBaseSerializer(
-        many=True, allow_empty=True, help_text="List of knowledge base items"
+    kbItems = serializers.ListField(
```
jgonggrijp: I think you could also make this `KnowledgeBaseAnnotationSerializer(many=True)` instead of a `ListField`. Not sure about the exact tradeoffs though.

frontend/src/app/annotate/annotation-input/problem-details/problem-labels/manage-labels-modal/manage-labels-modal.component.spec.ts
```diff
@@ -64,156 +82,194 @@ describe("ManageLabelsModalComponent", () => {
...
-                available.forEach(label => {
-                    expect(selectedIds.includes(label.id)).toBe(false);
-                });
-                done();
-            });
+        const text = component.getAttachedByText(annotation);
+
+        expect(text).toContain("you");
+        // Accept both US and non-US date formats (for GitHub Actions runners.)
+        expect(text).toMatch(/(March 15, 2024|15 March 2024)/);
```
jgonggrijp: This seems a bit flaky. Can you control for the locale in the unittest?

frontend/src/app/annotate/annotation-input/problem-details/problem-labels/manage-labels-modal/manage-labels-modal.component.spec.ts
```diff
@@ -64,156 +82,194 @@ describe("ManageLabelsModalComponent", () => {
...
+            attachedByCurrentUser: false,
+            createdBy: "Other User",
+            createdAt: "2024-03-15T10:30:00Z"
+        };
+
+        const text = component.getAttachedByText(annotation);
+
+        expect(text).toContain("Other User");
+        // Accept both US and non-US date formats (for GitHub Actions runners.)
+        expect(text).toMatch(/(March 15, 2024|15 March 2024)/);
```
jgonggrijp: These lines are almost the same as 202-206.

frontend/src/app/annotate/annotation-input/problem-details/problem-labels/manage-labels-modal/manage-labels-modal.component.spec.ts
```diff
@@ -64,156 +82,194 @@ describe("ManageLabelsModalComponent", () => {
...
-            let availableCount = 0;
-            component.availableLabels$.subscribe(labels => {
-                availableCount = labels.length;
-            });
-            tick();
+    it("should set loadingLabels$ to false after labels are available", fakeAsync(() => {
+        let loading = true;
+        component.loadingLabels$.subscribe(isLoading => {
+            loading = isLoading;
+        });
```
jgonggrijp: I also see this pattern a lot throughout the test module. This might be worth making a helper function for.

frontend/src/app/annotate/annotation-input/problem-details/problem-labels/manage-labels-modal/manage-labels-modal.component.ts
```diff
@@ -31,31 +30,70 @@ export class ManageLabelsModalComponent implements OnInit {
...
+                    session: null,
+                    removedAt: null,
+                    removedBy: null,
+                    notes: '',
+                    removable: true,
+
+                };
+                return newAnnotation;
+            }).filter((a) => a !== null);
+        })
```
jgonggrijp: please deforest

frontend/src/app/annotate/annotation-input/problem-details/problem-labels/manage-labels-modal/manage-labels-modal.component.ts
jgonggrijp, line 1: I see a lot of linear searching in this module, in some cases nested inside larger loops. Are the searched arrays guaranteed to be small? Otherwise, it might be worth using faster lookup structures.

frontend/src/app/util.ts
jgonggrijp, line 1: I think I saw tests in `manage-labels-modal.component.spec.ts` that were mostly testing the behavior of the functions in this module. It would be better to have those tests in a dedicated `util.spec.ts`, because that makes them easier to find when debugging these functions.

jgonggrijp (changes requested):
I have read the code with attention, but did not try out the feature.

There is a lot of code and I could not completely oversee the larger structure of it, but most of it looks reasonable to me. I made a relatively large number of comments, but they are all about technical details. I really appreciate the extensive testing of two of the modules.

> This PR builds on #63 and includes the latest version of develop. It develops the original PR's idea of using KnowledgeBaseAnnotations, bound by a AnnotationSession, instead of a KnowledgeBase model that is appended to a Problem directly. I think this is a good change which fits the researcher's vision that user annotations will mainly consist of additions/edits of knowledge base items and proof trees.

I suspect this `AnnotationSession` might end up being a temporary placeholder for what is essentially a proof. This works for now, though.

> AnnotationSessions also have a "notes" field aimed at remarks or justifications. I first implemented that too, but noticed that any notes given when adding a label are overwritten when a label is removed. This would have been easier to implement with the 'Event'/'Change' structure you proposed in an earlier PR review.

This reads a bit like an open end for a follow-up branch. What do you want to do in that follow-up? Drop the feature? Have another try at making it work? Switch to an event based data model? In the latter case, I don't think this branch should be merged as is.

backend/annotation/serializers.py
XanderVertegaal: I agree!

backend/annotation/models.py
XanderVertegaal: Snatched from the clutches of oblivion.

backend/annotation/serializers.py
XanderVertegaal: It was never here. *jedi hand wave*

frontend/src/app/annotate/annotation-input/problem-details/problem-labels/manage-labels-modal/manage-labels-modal.component.ts
XanderVertegaal: Chop-chop! I'm also switching to lookup maps that are created once to avoid nested loops.

frontend/src/app/annotate/annotation-input/problem-details/problem-labels/manage-labels-modal/manage-labels-modal.component.spec.ts
XanderVertegaal: Done!

Issue #19, opened 2025-05-14
 Display proof trees and syntactic parses in frontend


XanderVertegaal:
Wrangle proof trees / tableaux and design interface for the API.

- Use mock data to generate tree structure / XML
- Decide whether to keep XML or transform to another format (e.g. JSON)? Lasha could help with this. Preferably something we can type strongly.
- Bonus: fetch in frontend and display (could be separate issue)

kovvalsky:
I think this issue can be closed, right? @XanderVertegaal

PR #72, opened 2026-02-06
Feature/display syntactic parses

XanderVertegaal:
Addresses part of #19 

This PR introduces the ParseTreeTableComponent, which is used to display syntactic parses in the frontend.

image

Since the actual output of LangPro's syntactic parse is still unknown (to me), I have inductively constructed a set of types (in `types.ts`) based on the HTML of LangPro's current working demo. These types are then mapped to a 'display object' with its own types (`BaseCell`, `NodeCell` etc.) that is used in the template. Decoupling the parser output from the tabular display object in this way ensures that the table logic can stay intact when we (inevitably) have to adapt the type structure of the parser output based on LangPro's actual return values. 

Lastly, I have also included a `Highlight` directive that highlights any hovered cell and its children. The original LangPro demo has a similar feature.


frontend/src/app/annotate/annotation-parse-results/parse-tree-table/parse-tree-highlight.directive.ts
```diff
@@ -0,0 +1,103 @@
...
+    colEnd: number;
+}
+
+@Directive({
+    selector: '[laParseTreeHighlight]',
+    standalone: true
+})
+export class ParseTreeHighlightDirective implements AfterViewInit, OnDestroy {
+    private cellPositions: CellPosition[] = [];
+    private listeners: (() => void)[] = [];
```
jgonggrijp, line 18: Maybe add a comment to clarify that the listeners are actually functions to undo the listening. I suspected as much when I read your code with attention, but Angular's reference documentation for `Renderer2.listen` didn't clarify the purpose of the returned function. I had to consult Angular's source code to confirm.

frontend/src/app/annotate/annotation-parse-results/parse-tree-table/parse-tree-highlight.directive.ts
```diff
@@ -0,0 +1,103 @@
...
+    private initializeCellPositions(): void {
+        const table = this.el.nativeElement;
+        const rows = table.querySelectorAll('tbody > tr');
+
+        rows.forEach((row, rowIndex) => {
+            const cells = row.querySelectorAll('td');
+            let colStart = 0;
+
+            cells.forEach(cell => {
+                const colspan = parseInt(cell.getAttribute('colspan') || '1', 10);
```
jgonggrijp, line 43: It seems a little bit backwards that you're rendering a `colspan` somewhere and then have to tease out the value from the HTML again. Might be worth thinking how you can render it as a recursive component where each subtree just knows its own witdth. Not a requirement for acceptance, though.

frontend/src/app/annotate/annotation-parse-results/parse-tree-table/parse-tree-highlight.directive.ts
```diff
@@ -0,0 +1,103 @@
...
+            if (cellPos.rowIndex >= hoveredCell.rowIndex) {
+                // Cell is on a row below or the same as the hovered cell.
+                // Do not highlight.
+                return;
+            }
+            if (cellPos.colStart >= hoveredCell.colStart && cellPos.colEnd <= hoveredCell.colEnd) {
+                this.renderer.addClass(cellPos.element, 'highlight');
+            }
+        });
+    }
```
jgonggrijp, line 94: This is a lot of logic for just highlighting a piece of a table. With a recursive structure, where subtrees are child elements of the hovered tree, you could leave this to CSS.

frontend/src/app/annotate/annotation-parse-results/parse-tree-table/parse-tree-table.component.ts
```diff
@@ -0,0 +1,144 @@
...
+
+
+@Component({
+    selector: 'la-parse-tree-table',
+    imports: [SubscriptAngleBracketsPipe, ParseTreeHighlightDirective],
+    templateUrl: './parse-tree-table.component.html',
+    styleUrl: './parse-tree-table.component.scss'
+})
+export class ParseTreeTableComponent implements OnInit {
+    public readonly tree = input.required<ParseTree>();
```
jgonggrijp, line 20: What does `input.required` do? Is it related to/a new notation for `@Input`? I'm asking because the combination with `readonly` surprises me.

frontend/src/app/annotate/annotation-parse-results/parse-tree-table/parse-tree-table.component.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    public readonly tableRows: TableCell[][] = [];
+
+    public readonly treeTypeDisplay = signal("");
+
+    ngOnInit(): void {
+        const root = this.tree().root;
+        const rows = this.createTableRows(root);
+        this.tableRows.push(...rows);
```
jgonggrijp: Why not just direct initialize like below? It is important that the `tableRows` reference keeps pointing to the same object?


```suggestion
        this.tableRows = this.createTableRows(root);
```

frontend/src/app/annotate/annotation-parse-results/parse-tree-table/parse-tree-table.component.ts
```diff
@@ -0,0 +1,144 @@
...
+        const binaryTotalWidth = this.getRowWidth(combinedRows[0]);
+        combinedRows.push([{
+            content: node.cat,
+            rule: node.rule,
+            colspan: binaryTotalWidth,
+            type: "node"
+        }]);
+
+        return combinedRows;
+    }
```
jgonggrijp: This is again a lot of logic to calculate widths and offsets, that I think could be avoided by using a recursive structure.

jgonggrijp (commented):
Good job on finishing this so quickly!

The code looks reasonable to me. The logic is however quite complicated. I would prefer that this be addressed in one of the following ways (or possibly a bit of both):

- [ ] (good) verify that the logic is correct with unittests;
- [x] (better) remove complexity by rendering subtrees recursively. I just mean the structure under the hood; the table-like appearance is fine.

You don't have to address this concern before merging, but in that case I request that you document this in a ticket.

I also cloned the branch locally to preview the feature, and I **love** the way it looks, both with and without highlighting. This is something we can present with pride. 👍

frontend/src/app/annotate/annotation-parse-results/parse-tree-table/parse-tree-table.component.ts
XanderVertegaal: [`input()`](https://angular.dev/api/core/input#) is one of these new Angular signals, stable since Angular 19. The older [decorator syntax](https://angular.dev/api/core/Input) (`@Input`) is still supported though. Adding `.required()` means that TS will complain at build time if a parent component uses this component without providing a value for it. In that sense it is identical to `@Input({ required: True})` in the old syntax. The difference lies mainly in the way Angular watches for changes. (There is a move away from ZoneJS to zoneless change detection based on signals.)

I use `readonly` to have TS warn me about reassignments of property at compile time. I don't think it is a real risk, but I like it as a way of documenting that this property should never be mutated.

frontend/src/app/annotate/annotation-parse-results/parse-tree-table/parse-tree-highlight.directive.ts
XanderVertegaal: I would love to do this, but how would you implement this is if the rendered `<table>` is rendered with the root node at the bottom and the nested nodes at the top?

jgonggrijp: Preorder recursion.

XanderVertegaal:
I went for the 'better' option and now use a recursive 'TreeNode' component using nested `div`s instead of a table. The highlighting is now offloaded to the CSS, which is a lot better :)

I could not get left and right borders to work correctly, so am now only adding a top border for every non-terminal node. It gives a more minimalistic style, but it matches what is on the current LangPro demo.

frontend/src/app/annotate/annotation-parse-results/parse-tree-table/parse-tree-table.component.ts
```diff
@@ -38,28 +12,18 @@ const TreeTypeDisplay: Record<ParseTreeType, string> = {
...
-        const rows = this.createTableRows(root);
-        this.tableRows.push(...rows);
-    }
-
-    public getTreeTypeDisplay(type: ParseTreeType): string {
-        return TreeTypeDisplay[type] || "Unknown Type";
-    }
-
-    private createTableRows(node: ParseTreeNode): TableCell[][] {
+    private buildDisplayTree(node: ParseTreeNode): TreeNodeDisplay {
```
jgonggrijp, line 26: Tip: look up the Builder pattern in the GoF when you're at the lab (we have a copy there). Not a change request, just something that you may find inspiring for the next time you need to build something like this.

frontend/src/app/annotate/annotation-parse-results/parse-tree-table/parse-tree-table.component.ts
```diff
@@ -72,73 +36,52 @@ export class ParseTreeTableComponent implements OnInit {
...
-        }
-
-        // Combine rows horizontally
-        const combinedRows: TableCell[][] = [];
-        for (let i = 0; i < maxRows; i++) {
-            combinedRows.push([...leftRows[i], ...rightRows[i]]);
-        }
-
-        // Add parent node row
-        const binaryTotalWidth = this.getRowWidth(combinedRows[0]);
```
jgonggrijp, line 121: I'm so happy that all of this can go.

frontend/src/app/annotate/annotation-parse-results/parse-tree-table/tree-node.component.scss
```diff
@@ -0,0 +1,24 @@
...
+
+        .tree-node {
+            background-color: var(--bs-yellow);
+        }
+    }
+}
+
+:not(.terminal-node) > .node-content {
+    border-top: 1px solid var(--bs-gray-500);
+}
```
jgonggrijp, line 19: I *think* this might give you borders as you had them with the table, but I have not tested it.

```suggestion
.tree-node {
    transition: background-color 0.2s ease;
    min-height: 10em;
    text-align: center;
    border: 1px solid var(--bs-gray-500);
    
    .tree-node {
        border-top: none;
        border-right: none;
        &:first-child {
            border-left: none;
        }
    }

    // Only highlight the current hovered node
    // and its children, not its ancestors.
    &:hover:not(:has(.tree-node:hover)) {
        background-color: var(--bs-yellow);

        .tree-node {
            background-color: var(--bs-yellow);
        }
    }
}
```

jgonggrijp (approved):
Much tidier and more maintainable, well done!

PR #73, opened 2026-02-23
Feature/allow kb edits

XanderVertegaal:
This PR fixes a couple of issues revolving around KB items.

- Creating/editing/deleting KB items is no longer limited to problem belonging to the USER dataset. All problems (incl. those from SICK, FraCaS and SNLI datasets) can now have KB annotations.

- The "Add knowledge base item" and "Save problem" buttons are now visible if the user has the `annotation.changeknowledgebase` permission. Their visibility used to depend indirectly on `user.can_edit_problem`, which is meant for changes to a problem's hypothesis and sentences.

- KB items can now be deleted. This was not possible before.

### DIFF
--- a/backend/problem/serializers.py
+++ b/backend/problem/serializers.py
@@ -1,4 +1,5 @@
 from rest_framework import serializers
+from django.utils import timezone
 
 from annotation.serializers import KnowledgeBaseAnnotationSerializer
 from annotation.models import (
@@ -120,8 +121,7 @@ class ProblemInputSerializer(serializers.Serializer):
         problem.premises.set(premise_sentences)
 
         kb_items = validated_data.get("kbItems", [])
-        if kb_items:
-            self._create_update_kb_annotations(problem, kb_items)
+        self._handle_kb_annotations(problem, kb_items)
 
         return problem
 
@@ -157,20 +157,39 @@ class ProblemInputSerializer(serializers.Serializer):
         serializer.is_valid(raise_exception=True)
         serializer.save(problem=problem, session=session, created_by=session.user)
 
-    def _create_update_kb_annotations(
+    def _mark_kb_not_in_input_as_removed(
+        self, problem: Problem, kb_items: list[dict], session: AnnotationSession
+    ) -> None:
+        """
+        Marks KnowledgeBase annotations for a problem that are not included in
+        the provided list of kb_items as removed.
+        """
+        kb_item_ids = {kb_item.get("id") for kb_item in kb_items if kb_item.get("id") is not None}
+
+        annotations_to_delete = KnowledgeBaseAnnotation.objects.filter(
+            problem=problem,
+            removed_at__isnull=True
+        ).exclude(id__in=kb_item_ids)
+
+        for annotation in annotations_to_delete:
+            annotation.removed_at = timezone.now()
+            annotation.removed_by = session.user
+            annotation.save()
+
+    def _handle_kb_annotations(
         self, problem: Problem, kb_items: list[dict]
     ) -> None:
         """
-        Creates or update KnowledgeBase and Label annotations for a problem.
+        Creates, updates and deletes KnowledgeBase annotations for a problem.
         Creates an annotation session if it does not exist.
-
-        TODO: handle deletions!
         """
         request = self.context.get("request", None)
-        if not request or not request.user.is_authenticated:
+        if not request or not request.user.is_authenticated or not request.user.can_edit_kb:
             return
 
         session = AnnotationSession.objects.create(user=request.user)
+
+        self._mark_kb_not_in_input_as_removed(problem, kb_items, session)
 
         for kb_item in kb_items:
             self._create_update_kb_annotation(kb_item, problem, session)
@@ -183,8 +202,7 @@ class ProblemInputSerializer(serializers.Serializer):
 
         # KB annotations can be made for all problems.
         kb_items = validated_data.get("kbItems", [])
-        if kb_items:
-            self._create_update_kb_annotations(instance, kb_items)
+        self._handle_kb_annotations(instance, kb_items)
 
         # Other fields can only be updated for user-created problems.
         if instance.dataset != Problem.Dataset.USER:
--- a/backend/problem/serializers.py
+++ b/backend/problem/serializers.py
@@ -77,11 +77,9 @@ class ProblemInputSerializer(serializers.Serializer):
     base = serializers.IntegerField(required=False, allow_null=True)
 
     def validate_id(self, value):
-        """Validate that the Problem ID, if provided, exists and belongs to a user-created problem."""
+        """Validate that the Problem ID, if provided, exists."""
         if value is not None:
-            if not Problem.objects.filter(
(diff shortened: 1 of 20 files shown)


backend/problem/serializers.py
```diff
@@ -159,21 +157,40 @@ def _create_update_kb_annotation(
...
+        """
+        kb_item_ids = {kb_item.get("id") for kb_item in kb_items if kb_item.get("id") is not None}
+
+        annotations_to_delete = KnowledgeBaseAnnotation.objects.filter(
+            problem=problem,
+            removed_at__isnull=True
+        ).exclude(id__in=kb_item_ids)
+
+        for annotation in annotations_to_delete:
+            annotation.removed_at = timezone.now()
```
jgonggrijp: Maybe evaluate `timezone.now()` once before the loop and then reuse the same value inside the loop. As it is now, you could get one-second timestamp differences between annotations that were deleted together, which might be confusing for human observers.

jgonggrijp: Actually, what kind of differences will be observed depends on rounding. With coarser rounding, differences become less likely but more confusing.

backend/problem/serializers_test.py
```diff
@@ -131,3 +113,250 @@ def test_blank_hypothesis_invalid():
...
+            "relationship": kb_annotation.relationship,
+        }
+    ]
+
+    serializer = input_serializer_with_user(visitor)
+    serializer._handle_kb_annotations(user_problem, kb_input) # type: ignore
+
+    # Verify KB annotation was not updated
+    kb_annotation.refresh_from_db()
+    assert kb_annotation.entity1 == original_entity1, "KB annotation should not have been modified without permission."
```
jgonggrijp:
```suggestion
    assert kb_annotation.entity1 == original_entity1, "KB annotation should not have been modified without permission."
    assert kb_input[0]['entity1'] != original_entity1
```

Check your assumptions!

backend/problem/serializers_test.py
```diff
@@ -131,3 +113,250 @@ def test_blank_hypothesis_invalid():
...
+        }
+    ]
+
+    serializer = input_serializer_with_user(annotator)
+    serializer._handle_kb_annotations(user_problem, kb_input)
+
+    kb_annotations = KnowledgeBaseAnnotation.objects.filter(
+        problem=user_problem, removed_at__isnull=True
+    )
+    assert kb_annotations.count() == 1, "One KB annotation should have been created."
```
jgonggrijp: The assertion would also pass if `serializer._handle_kb_annotations` created two or more annotations and already marked all except for one as removed before returning. To rule that out, you'd have to remove the `removed_at__isnull` filter criterion.

backend/problem/serializers_test.py
```diff
@@ -131,3 +113,250 @@ def test_blank_hypothesis_invalid():
...
+    assert kb_annotation.entity2 == "updated_e2"
+    assert kb_annotation.relationship == "not_equal"
+
+    # Verify the new annotations
+    new_annotations = kb_annotations.exclude(id=kb_annotation.pk)
+    assert new_annotations.count() == 2, "There should be two new KB annotations after update."
+
+    entities = [(kb.entity1, kb.entity2) for kb in new_annotations]
+    assert ("new_e1", "new_e2") in entities
+    assert ("another_e1", "another_e2") in entities
```
jgonggrijp, line 313: Nice approach.

backend/problem/serializers_test.py
jgonggrijp, line 1: I like the thoroughness and the combinatoric coverage of the new tests in this module. However, it stands out to me that they all test the private `_handle_kb_annotations` method rather than the public `create` or `update` method that would most likely call it. If `_handle_kb_annotations` works correctly but some other part of `create` or `update` doesn't, then the public method as a whole could still malfunction in principle, even with regard to adding, editing or remove-marking annotations.

I'm just pointing this out as a weak spot in the test coverage; I don't have a ready opinion on how to address it. Testing a private method in itself isn't wrong, and I appreciate your approach as a unified way to test the common functionality between `create` and `update`. I have a few ideas, you might have more:

- Leave some comments in the source code of `create` and `update` to remind future editors that logic with regard to adding, modifying and remove-marking annotations belongs in the `_handle_kb_annotations` method.
- Add a few tests (not as thorough as for `_handle_kb_annotations`) that ensure that the behavior of `create` and `update` doesn't deviate from the private method. Maybe a fuzz test that tries different random preconditions and random input each time and just verifies that both the public and the private method handle the annotations in the same way?
- Document as an issue (perhaps a bit "heavy" for what it is).
- Accept the uncertainty and do nothing. Not my favorite option, I somewhat expect this to come back and bite us after we've forgotten the details, but you may be able to convince me.

frontend/src/app/annotate/annotation-input/knowledge-base-form/knowledge-base-form.component.html
```diff
@@ -26,7 +25,7 @@
         `${i}-${kbItem.value.entity1}-${kbItem.value.relationship}-${kbItem.value.entity2}`)
         {
         <li class="mb-3">
-            @if (appMode === 'add' || appMode === 'edit') {
+            @if (canEditKb$ | async) {
```
jgonggrijp: Question out of curiosity: is there any performance penalty to having the same observable piped through `async` in two places in the same template? I'm not expecting it to be a problem here, just asking to learn for future usage.

frontend/src/app/annotate/annotation-input/annotation-input.component.html
```diff
@@ -15,7 +15,7 @@
             <la-icon-button
                 buttonStyle="success"
                 [icon]="faTree"
-                label="Start parse"
+                label="Parse and proof"
```
jgonggrijp: Nouns or verbs?

jgonggrijp (commented):
Looks good to me. I have some comments for your perusal.

frontend/src/app/annotate/annotation-input/knowledge-base-form/knowledge-base-form.component.html
XanderVertegaal: I had to look this up myself, and every `| async` call sets up a new subscription. If the observable is not shared (as we do in the case of `canEditKb$` with `shareReplay`), I think you will get a new request to the backend for each individual subscription.

With Angular 19+, this risk of multiple backend requests can probably also be avoided by using only one `async` pipe in combination with `@let`:

```ts
@let canEdit = canEditKb$ | async;
```

Good to keep in mind; thanks for pointing this out!

frontend/src/app/annotate/annotation-input/annotation-input.component.html
XanderVertegaal: I'm too invested in bread-making!

(I'm changing it to 'Parse and prove' now.)

backend/problem/serializers_test.py
XanderVertegaal: I see your point. I think your second option is best and ensures full coverage. I'll see if I can add a few broad-stroke tests for the public methods.

PR #75, opened 2026-02-27
Refactor: permission checks in view

XanderVertegaal:
This PR updates the permission checks in the backend and lays some groundwork for role-based UI differentiation (frontend changes) featured in #76.

The main change here is that permission checking is moved out of the ProblemInputSerializer and into the ProblemView. There are a few advantages to this. First of all, the ProblemInputSerializer can focus on serializing, i.e. creating, updating and removing Problems and KB annotations based on input data. Secondly, the serializer no longer needs the request or the user to be passed when we initialize it, so it has one less dependency and we do not have to worry about forgetting to provide it. As a result, the serializer tests have been simplified (merely checking correct CRUD operations), and the view-related tests have been expanded accordingly. They take the role of the broader checks implemented in the last PR, to make sure requests change the database in the expected way.

I think this is a more robust situation that is closer to what views and serializers should be doing.

As a bonus, the permission matrix in the README is updated.


backend/problem/views/problem_test.py
```diff
@@ -145,30 +154,255 @@ def test_visitor_cannot_update_problem(
...
+
         client.force_login(user=annotator)
         response = client.patch(
             f"/api/problem/{sample_problem.id}/",
             problem_input_data,
             content_type="application/json",
         )
+
+        # Status code 200 because KB items are updated in the same view.
         assert response.status_code == status.HTTP_200_OK
```
jgonggrijp: Please clarify

backend/problem/views/problem_test.py
```diff
@@ -145,30 +154,255 @@ def test_visitor_cannot_update_problem(
...
+        assert (
+            kb_annotation.entity1 == problem_input_data["kbItems"][0]["entity1"]
+        ), "KB annotation entity1 should match input data."
+
+    def test_annotator_can_update_kb_annotations(
+        self, client, annotator, sample_problem, problem_input_data, kb_annotation
+    ):
+        assert sample_problem.knowledgebaseannotations.count() == 1
+        first_kb = sample_problem.knowledgebaseannotations.first()
+        assert first_kb.entity1 != problem_input_data["kbItems"][0]["entity1"]
```
jgonggrijp, line 290: I love how you're doing this everywhere now.

backend/problem/views/problem_test.py
jgonggrijp, line 1: These are very good, thorough tests.

jgonggrijp (approved):
Yes, this does look cleaner. Good work!

backend/problem/views/problem_test.py
XanderVertegaal: In this test, the user attempts to update core problem fields, but the user is not allowed to do so. In such cases, you might expect 403 FORBIDDEN to be returned.

The user _is_ allowed to update KB items, however, so that part of the request is still carried out. In that sense the request is successful and that is why 200 OK is returned.

I'll update the comment to clarify this.

backend/problem/views/problem_test.py
XanderVertegaal: I kept saying "Check your assumptions!" from your previous review to myself when writing these tests 😆 It is a good habit to get into.

backend/problem/views/problem_test.py
jgonggrijp: Hm, in that case I think I disagree with the approach. If part of a request is allowed and part of it is not, I think the entire request should be rejected. Returning a 200 is potentially confusing as well. Let the client come back when they have better permissions or a more modest request.

XanderVertegaal: Returning a 403 (or another error code) is potentially confusing as well, since the request was not rejected as a whole. I could not find a good answer for this, other than [https://stackoverflow.com/questions/8472935/http-status-code-for-a-partial-successful-request](non-standard status codes) such as [202 ACCEPTED](https://datatracker.ietf.org/doc/html/rfc7231#section-6.3.3), which seems more suitable for asynchronous requests.

Maybe the best way is to inspect the request to see if the user has indeed submitted new values for any core problem fields and return 403 if that is the case, or 200 otherwise. How does that sound?

jgonggrijp: Yes that sounds reasonable. That was what I intended, I think.

PR #76, opened 2026-02-27
Feature/frontend permissions

XanderVertegaal:
This PR makes sure that the frontend allows users to do what they are allowed to, according to the roles and permissions in the permission matrix (see README.md in #75).

- User has two new computed properties: `can_copy_problem` and `can_add_label_annotations` to govern button visibility in the frontend.
- LabelAnnotation management relies on `add_labelannotation` only, and the `change_labelannotation` permission is removed. There is no difference between the two, so this felt simpler, but let me know if you disagree or think I should rename `add_labelannotation` instead.
- The text you see when hovering over a label used to be computed in very similar ways in two places. This has been outfactored to a single function. Also, this text would show you the ID of the annotator rather than their name. This has been fixed.

### DIFF
--- a/backend/annotation/serializers.py
+++ b/backend/annotation/serializers.py
@@ -17,7 +17,7 @@ class AnnotationBaseSerializer(serializers.ModelSerializer):
     """
 
     createdAt = serializers.DateTimeField(source="created_at", read_only=True)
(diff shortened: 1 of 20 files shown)


jgonggrijp (approved):
Looks good to me!

PR #77, opened 2026-03-06
Integration with parser backend

jgonggrijp:
This branch continues and supersedes the work in #51. Merging this PR will also merge #51 as a side effect.

The changes in this branch depend on https://github.com/CentreForDigitalHumanities/langpro-container/pull/4. The langpro-container is added as a git submodule.

As discussed last Tuesday with @XanderVertegaal and @kovvalsky, the container has been slightly adapted to also respond with JSONified nltk.Tree-ified proofs in non-raw mode. The JSONified nltk.Trees lack labeling of tree components; all structure beyond basic subtree nesting is implicitly encoded in the string label of each node. We will initially work with the raw node labels without parsing them, which in some places may require simplifying the visualizations. As further discussed with @XanderVertegaal, this branch does not yet attempt to actually render these data, but merely ensures that an actual request is made to the container and that its response is delivered to the right components.

The `frontend/src/app/tree.ts` and all modules under `frontend/src/app/annotate/parse-tree`, which were added in #51, are now effectively dead code and should be removed eventually. You can safely ignore them in review.

I have not made any effort yet to pipe the response to the proof tableau components; I will do this in a follow-up branch.

Please review this in whatever way makes the most sense to you. I will comment on a few details in advance.


frontend/src/app/annotate/annotation-parse-results/annotation-parse-results.component.ts
```diff
@@ -11,6 +13,19 @@ import { NgbAccordionModule } from "@ng-bootstrap/ng-bootstrap";
     templateUrl: "./annotation-parse-results.component.html",
     styleUrl: "./annotation-parse-results.component.scss",
 })
-export class AnnotationParseResultsComponent {
-    public readonly parseResults = mockResult;
+export class AnnotationParseResultsComponent implements OnInit {
+    private destroyRef = inject(DestroyRef);
+    private parseService = inject(ParseService);
+
+    public parseResults = mockResult;
```
jgonggrijp, line 20: I left the initialization with the `mockResult` for now, because it provides a simple visual confirmation that the response from the container is actually fed to this component. When you press the "Parse and prove" button, the parse tables disappear (with a delay because the response arrives with a delay). This is because the CCG trees are in a format that the component has not been adapted to yet.

frontend/src/app/annotate/annotation-parse-results/annotation-parse-results.component.ts
```diff
@@ -11,6 +13,19 @@ import { NgbAccordionModule } from "@ng-bootstrap/ng-bootstrap";
...
+
+    ngOnInit(): void {
+        // Subscription needed to ensure a request is actually made.
+        this.parseService.parse$
+            .pipe(takeUntilDestroyed(this.destroyRef))
+            .subscribe((response) => {
+                console.log("Parse response:", response);
+                this.parseResults = response.data.ccg_trees;
+            });
```
jgonggrijp, line 29: It would probably be better to do something like

```suggestion
        this.parseResults$ = this.parseService.parse$
            .pipe(takeUntilDestroyed(this.destroyRef))
            .pipe(map((response) => response.data.ccg_trees));
```

and then write `@let parseResults = parseResults$ | async` in the template. I leave this up to you, when you adapt the component to the actual format of the data.

frontend/src/app/annotate/annotation-parse-results/annotation-parse-results.component.ts
```diff
@@ -1,5 +1,7 @@
-import { Component } from "@angular/core";
+import { Component, DestroyRef, inject, OnInit } from "@angular/core";
+import { takeUntilDestroyed } from "@angular/core/rxjs-interop";
 import { mockResult } from "./mockParseResult";
+import { ParseResponse, ParseService } from "@/services/parse.service";
```
XanderVertegaal: Unused import.
```suggestion
import { ParseService } from "@/services/parse.service";
```

frontend/src/app/annotate/annotate.component.ts
```diff
@@ -1,7 +1,7 @@
 import { Component, DestroyRef, inject, OnInit } from "@angular/core";
 import { AnnotationMenuComponent } from "./annotation-menu/annotation-menu.component";
 import { NavigatorComponent } from "./navigator/navigator.component";
-import { AnnotationInputComponent } from "./annotation-input/annotation-input.component";
+import { AnnotationInputComponent, ParseInput } from "./annotation-input/annotation-input.component";
```
XanderVertegaal:
```suggestion
import { AnnotationInputComponent } from "./annotation-input/annotation-input.component";
```

frontend/src/app/services/parse.service.ts
```diff
@@ -3,6 +3,8 @@ import { inject, Injectable } from '@angular/core';
 import { Subject, switchMap, catchError, of } from 'rxjs';
 import { ParseInput } from '@/annotate/annotation-input/annotation-input.component';
 
+export type ParseResponse = any;
```
XanderVertegaal, line 6: I assume that this is a placeholder for when we actually define the expected output from the parser (in a future PR).

frontend/package.json
```diff
@@ -37,6 +37,7 @@
     "colors": "^1.4.0",
     "express": "^4.18.2",
     "rxjs": "~7.8.0",
+    "svg-pan-zoom": "bumbu/svg-pan-zoom",
```
XanderVertegaal, line 40: Is there a reason to refer to the Git repo instead of the NPM registry (with e.g. `"svg-pan-zoom": "^3.6.2"`) for this package?

XanderVertegaal: Ah, I should actually be asking @bbonf about this, I think :)

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "langpro-container"]
+	path = langpro-container
+	url = git@github.com:CentreForDigitalHumanities/langpro-container.git
```
XanderVertegaal: The SSH approach did not work with me. (I got permission errors with my public key.) Using HTTPS as follows fixed the issue.

```
	url = https://github.com/CentreForDigitalHumanities/langpro-container.git
```

Should I look into fixing my SSH public key, or can we adopt HTTPS?

XanderVertegaal (approved):
The code looks fine, and I'm happy you got everything going!

Locally, I'm running into `ModuleNotFoundErrors`, but I will contact you separate about this. I suspect this has to do with my local setup, not with the code changes proposed in this PR.

.gitmodules
jgonggrijp: I think you need to ask @tymees for help. If I'm right, the repository now belongs to our team. I'm able to sync using SSH so in theory, so should you.

frontend/src/app/services/parse.service.ts
jgonggrijp: This is what @bbonf originally wrote. I made the same assumption as you did.

.gitmodules
jgonggrijp: Actually, I now realize that the SSH URL might lead to problems in CI and deployment. I'll change it to https.

jgonggrijp: @XanderVertegaal If you actually have problems with your SSH key, I of course still recommend that you address those. SSH is easier for syncing because you don't need to reauthenticate periodically.

PR #78, opened 2026-03-16
Feature/langpro parse results

XanderVertegaal:
This PR displays syntactic parse results in the "CCG/LLF" tab of the Annotator interface. 

The backend server now sends serialized results for all four parse trees to the frontend, but only the CCG Tree is currently visualised. The other three (CCG Term, Corrected CCG Term and LLF) are transformed into an NLTK in a different way (using `parse_term` instead of `parse_ccg_tree`, as indicated by @kovvalsky). This yields malformed visualisations in the frontend, which relies on the CCG Tree format.

We can of course handle these two types of trees differently in the frontend, but in my opinion it makes more sense to try to arrive at a common tree output format for both CCG Tree and the other three. This is something to discuss during our next meeting.

As a bonus, this PR enables scrolling in the ParseTreeTable, so complicated sentence parses (such as SICK 1) can be shown in full without overflowing the page.


frontend/src/app/annotate/annotation-parse-results/annotation-parse-results.component.ts
```diff
@@ -1,31 +1,51 @@
...
+    type: string;
+    tree: CCGNode;
+}
+
+interface FlattenedParseResult {
+    sentence: string;
+    ccgTrees: TreeWithType[];
+}
+
+function flattenResult(parse: CCGParse): FlattenedParseResult {
```
XanderVertegaal: "flattening" is probably not the most appropriate term here. Suggestions are welcome!

kovvalsky:
I agree with having a common tree output format for both CCG Tree and the other three terms.
The only difference between trees and terms is that trees require rule names as decorations of the horizontal lines because trees use several rules. Terms don't require the rule names to be displayed. So as long as the rule name element takes no space when its value is empty, the tree format can be used for terms, too.

frontend/src/app/annotate/annotation-parse-results/annotation-parse-results.component.html
```diff
@@ -20,6 +20,10 @@ <h2 ngbAccordionHeader>
                 </div>
             </div>
         </div>
+        } @empty {
+        <div class="mt-3">
+            <i class="text-muted" i18n>No parse results to display yet.</i>
```
jgonggrijp: As a help to the user, perhaps write something like "press the parse and prove button to see the parse results".

frontend/src/app/annotate/annotation-parse-results/parse-tree-table/parse-tree-table.component.ts
```diff
@@ -1,87 +1,115 @@
...
+    return {
+        type: 'leaf',
+        content: cat,
+        children: [],
+        leaf: { tok, lem, pos, ner }
+    };
+}
+
+function buildBinaryNode(node: BinaryNode): TreeNodeDisplay {
+    const left = buildDisplayTree(node.children[0]);
```
jgonggrijp, line 46: For very large parses, recursion could lead to a stack overflow. Not something to worry about right now, but useful to know.

frontend/src/app/annotate/annotation-parse-results/annotation-parse-results.component.ts
jgonggrijp: Lift? Collapse? Unfold?

frontend/src/app/annotate/annotation-parse-results/types.ts
jgonggrijp, line 1: The types in this module belong in a central place, because the `ParseService` and also the `TableauSVG` component will have to refer to at least some of these types.

jgonggrijp (changes requested):
This looks good to me. The only thing I'd ask you to change, is to move the types from the parse results component folder to a central place, possibly just `@/types`.

frontend/src/app/annotate/annotation-parse-results/parse-tree-table/parse-tree-table.component.ts
XanderVertegaal: Don't worry about it. Kubernetes will solve all of our problems.

(Just kidding; good point! Let's hope we never have to worry about it.)

frontend/src/app/annotate/annotation-parse-results/annotation-parse-results.component.ts
XanderVertegaal: Unfold it is!

PR #79, opened 2026-03-17
Visualize the actual proofs of the current problem with the TableauSVG component

jgonggrijp:
Together with #78 and #77, I think this finally completes the work that started with #51.

The pre-existing "Tableau / Proof" tab is split into a "Proof / Entailment" and a "Proof / Contradiction" tab, which visualize the respective proofs for the current problem. Please read the commit messages for a quick summary of my approach.

The tests pass and it generally seems to work well. That being said, at the time of writing, there are still a few known problems with the code on this branch. More on this below.

For now, I suggest trying the branch as follows.

1. Enter and save two or more small problems. If you don't already have a few, I suggest the following.

*The default example from the old demo*

```
Every man is working
Everybody who is working has an expensive car
---
Every man owns a car
```

*A similar example that I just came up with*

```
Every child likes to play
Everyone who likes to play is creative
---
Every child is creative
```

2. Navigate to one of these problems and make sure the "CCG / LLF" tab is selected.
3. Press the "parse and prove" button. Wait for the contents of the "CCG / LLF" tab to disappear.
4. Open the "Proof / Entailment" tab and observe a proof tableau.
5. Open the "Proof / Contradiction" tab and observe a different proof tableau.
6. Repeat from 2 with another small problem and observe two yet different proof tableaux.

It works for large problems, too, but in most cases, the overflow on the tab content element is preventing the proof from being visible. For example, if you try the first SICK problem, the Entailment tab will look empty, but if you inspect the DOM, you will find that the tableau is actually there. There is still a visible visualization in the Contradiction tab, but it is truncated on the right. I have held off from fixing this because I already took a quick peek at #78 and noticed that @XanderVertegaal fixed an overflow problem on that brach, which is probably the same issue.

Other known issues that I think are also best solved in coordination with #78:

- [ ] As mentioned above, the "CCG / LLF" tab must be selected when the "Parse and prove" button is pressed, in order for the tableaux to be rendered. This is because `AnnotationMenuComponent`, which passes the proof trees to `TableauSVG`, is not (yet) subscribed to the `AnnotationService#proofs$` observable. The `AnnotationsParseResultsComponent` is subscribed to the underlying `#parse$` observable, however, and is therefore used (for now) to trigger the request to the backend. We *could* subscribe in two places, but it's probably more elegant (and perhaps also more efficient?) to coordinate a single place where this is done.
- [ ] The types are just `any` all over the place. Without having seen the code, I suspect @XanderVertegaal already (partially) addressed this in #78.

Other other known issues:

- [ ] The `nltk2tableau` conversion function deserves some solid documentation. I still want to add this before merging.
- [ ] The `TableauNodeComponent` does not take the rule into account when calculating the width of leaf nodes (or of non-leaf nodes with very slim terms, for that matter). Also something I still want to fix before merging.

Without the still to be added comments, I expect the conversion logic to be difficult to follow if not outright mysterious. That being said, you are welcome to review it anyway and let me know what you find most opaque/offensive. :wink:


frontend/src/app/annotate/annotation-menu/annotation-menu.component.html
```diff
@@ -15,13 +17,30 @@
     <li [ngbNavItem]="2">
         <button ngbNavLink>
             <fa-icon [icon]="faTree" class="me-2" aria-hidden="true" />
-            <span i18n>Tableau / Proof</span>
+            <span i18n>Proof / Entailment</span>
```
XanderVertegaal, line 20: Much better! 'Tableau' is a representation but does not give information as to the actual content. (I don't call the syntactic parses "Table" either :P)

frontend/src/app/annotate/annotation-menu/annotation-menu.component.html
```diff
@@ -15,13 +17,30 @@
...
         </button>
         <ng-template ngbNavContent>
-            <la-annotation-tableau />
+            <section>
+                @if (proofs) {
```
XanderVertegaal, line 24: Perhaps for later, but maybe you could add an `@else` block with something like "No entailment data available yet."

frontend/src/app/annotate/annotation-menu/annotation-menu.component.html
```diff
@@ -15,13 +17,30 @@
...
+                <la-tableau-svg [tree]="proofs.entailment"></la-tableau-svg>
+                }
+            </section>
         </ng-template>
     </li>
     <li [ngbNavItem]="3">
+        <button ngbNavLink>
+            <fa-icon [icon]="faTree" class="me-2" aria-hidden="true" />
```
XanderVertegaal: I tried looking for another icon we could use to differentiate the entailment tree from the contradiction tree, but no obvious candidates present themselves. Maybe `link` and `linkSlash` for entailment and contradiction, respectively?

(This is obviously not a high-priority item. 😅)

frontend/src/app/services/parse.service.spec.ts
```diff
@@ -17,3 +20,115 @@ describe('ParseService', () => {
...
+                nodes: [{end: true, rule: 'closing_rule([1])'}],
+            }],
+        };
+        expect(nltk2tableau(nltk)).toEqual(tableau);
+    });
+
+    it('does not cause a stack overflow', () => {
+        let nltk: any = {node: 'Model'};
+        const tableau: any = {nodes: []};
+        for (let i = 0; i < 100000; ++i) {
```
XanderVertegaal, line 117: TIL about post-increments and pre-increments! Neat!

XanderVertegaal:
> The types are just any all over the place. Without having seen the code, I suspect @XanderVertegaal already (partially) addressed this in https://github.com/CentreForDigitalHumanities/langpro-annotator/pull/78.

I hate to disappoint, but I merely replaced `any` with `unknown` when it comes to the proofs. 😂

```ts
export type ParseResponseData = {
    ccg_parses: CCGParse[];
    proofs: unknown[];
};
```

If you need "any" help (see what I did there?) getting rid of the `any`s, let me know. I actually enjoy type wrangling.

frontend/src/app/annotate/tableau-svg/tableau-tree.component.ts
```diff
@@ -1,12 +1,12 @@
...
-import { TableauTerm } from './tableau-term.component';
+import { TableauNode } from './tableau-node.component';
 import { Dimensions } from '@/types';
 import { sum } from "@/util";
 
 @Component({
     selector: "[tableau-tree]",
     standalone: true,
-    imports: [TableauTerm],
+    imports: [TableauNode],
```
XanderVertegaal, line 9: My IDE says that this import is unused. Is that true?

frontend/src/app/services/parse.service.ts
```diff
@@ -23,4 +23,71 @@ export class ParseService {
             )
         )
```
XanderVertegaal: Two POST requests are being sent when 'Parse and prove' is clicked, presumably because this observable is subscribed to multiple times. If I add `share()` here, the requests are deduped.
```suggestion
        ),
        share()
```

frontend/src/app/services/parse.service.ts
```diff
@@ -23,4 +23,71 @@ export class ParseService {
...
+
+function extractProofs(response: ParseResponse) {
+    const {entailment, contradiction} = response.data.proofs;
+    return {
+        entailment: nltk2tableau(entailment),
+        contradiction: nltk2tableau(contradiction),
+    };
+}
+
+const emptyTableau = (): any => ({nodes: []});
```
XanderVertegaal: Out of curiosity: any reason why you opt for an arrow function instead of a traditional one here? Brevity?

frontend/src/app/services/parse.service.ts
```diff
@@ -23,4 +23,71 @@ export class ParseService {
             )
         )
     );
+
+    public proofs$ = this.parse$.pipe(map(extractProofs));
+}
+
+function extractProofs(response: ParseResponse) {
```
XanderVertegaal: Just a word of caution: `catchError` on line 21 returns `null`, so destructuring on line 32 will fail if that is the case. (This will undoubtedly be caught once we add proper typing.)

frontend/src/app/services/parse.service.ts
```diff
@@ -23,4 +23,71 @@ export class ParseService {
...
+    while (todo.length) {
+        // @ts-ignore
+        let task: [any, any] = todo.pop();
+        let newTasks = nltkNode2tableauNode(...task);
+        todo.push(...newTasks);
+    }
+    return tree;
+}
+
+function nltkNode2tableauNode(nltk: any, tree: any) {
```
XanderVertegaal: I await the documentation for this with bated breath! It feels very magical but at the same time there is not a lot of code, which is good.

XanderVertegaal (commented):
Good work so far! I'm glad you added the comment about the proof being there (but hidden out of screen) for some of the longer problems. I was very confused. 😅 My solution for wide output was simply to add overflow: scroll to a couple of containers, so there is space to scroll horizontally. Still, I wonder if this is the neatest and most 'overzichtelijke' solution. Perhaps adding an external library with a zoom and drag interface would be a good addition at some point for both the tableaus and the (often very wide) syntactic parse result tables.

One note: I could not get this branch to work without commenting out `this.parseResults = response.data.ccg_trees;` in the AnnotationParseResults component. (This should be fixed once both of our branches are merged.)

As indicated in one of the comments. Let me know if you would like to delegate the type wrangling to me. I will continue my review after you have had time to add documentation to the tree2Tableau magic! 🧙‍♂️

frontend/src/app/annotate/annotation-menu/annotation-menu.component.html
jgonggrijp: Maybe we should factor that out from the `AnnotationParseResults` component in #78 and reuse it across tabs to make them look consistent.

frontend/src/app/annotate/annotation-menu/annotation-menu.component.html
jgonggrijp: There are symbols in Unicode that exactly convey the intended meaning: ⊤ and ⊥. I went with those.

frontend/src/app/annotate/tableau-svg/tableau-tree.component.ts
jgonggrijp: The template of the tree component inserts instances of the node component. Maybe your IDE is failing to recognize that because the template is SVG instead of HTML?

frontend/src/app/services/parse.service.ts
jgonggrijp: Yes, if a function is just a single expression that fits on one line, I like to write it as an arrow.

frontend/src/app/services/parse.service.ts
jgonggrijp: Thanks for pointing that out to me! I just assumed that it would somehow magically prevent errors from being problematic in any way.

jgonggrijp:
@XanderVertegaal I responded to most of your comments so far and added comments for the `nltk2tableau` jugglefest. Please feel welcome to review that part of the code while I finalize the remaining work.

Here is my full plan. Please let me know if you think I'm missing something or if you'd prefer me to proceed differently.

1. Handle null response in `extractProofs`.
2. Use open/closed lock symbols for leaf nodes.
3. Make the tableau scroll on overflow.
4. Horizontally center proof tab on root node.
5. Take rule into account when calculating node widths. *I intend to get at least this far tomorrow.*
6. Wait for you to merge #78 (it's OK if you do this later than tomorrow).
7. Rebase on `develop`.
8. Add "real" types for the proof representations to the types that you wrote for #78.
9. Invite you for another review.
10. Address the "no trigger unless viewing parse tab" problem, if that still persists after the rebase.
11. Merge.

XanderVertegaal:
Your plan corresponds well to what we agreed, so please go ahead with the implementation. The only item on the list I would like you to clarify is 10: what do you mean with the 'no trigger unless viewing parse tab'?

frontend/src/app/annotate/annotation-menu/annotation-menu.component.html
XanderVertegaal: https://github.com/CentreForDigitalHumanities/langpro-annotator/issues/82

frontend/src/app/annotate/annotation-menu/annotation-menu.component.html
```diff
@@ -14,14 +16,31 @@
...
     <li [ngbNavItem]="2">
         <button ngbNavLink>
-            <fa-icon [icon]="faTree" class="me-2" aria-hidden="true" />
-            <span i18n>Tableau / Proof</span>
+            <b class="me-2" aria-hidden>⊤</b>
+            <span i18n>Proof / Entailment</span>
         </button>
         <ng-template ngbNavContent>
-            <la-annotation-tableau />
+            <section style="overflow-x: scroll;">
```
XanderVertegaal: No functional difference, but relying on the Bootstrap class keeps this code more maintainable.


```suggestion
            <section class="overflow-x-scroll">
```

frontend/src/app/annotate/tableau-svg/tableau-node.component.svg
```diff
@@ -14,8 +14,8 @@
 }
 
 @if(end) {
-    <svg:circle [attr.cx]="headX + 5" [attr.cy]="12" r="10" fill="salmon"/>
-    <svg:text #headText [attr.x]="headX" y="12" dominant-baseline="middle" font-weight="bold" fill="white">⛌</svg:text>
+    <svg:circle [attr.cx]="headX + 8" [attr.cy]="11" r="10" fill="salmon"/>
```
XanderVertegaal: I like the new symbols, but they are sadly still not aligned on my screen:

Image

Increasing the value of `[attr.cx]` seems to help, with `totalW/2` getting it close to perfect.

frontend/src/app/annotate/tableau-svg/tableau-svg.component.ts
```diff
@@ -25,6 +28,10 @@ export class TableauSVG {
     ngAfterViewChecked() {
         this.treeDimensions$.next(this.treeDimensions);
         this.cdref.detectChanges();
+        const width = this.treeDimensions.width;
+        if (width) this.element.nativeElement.parentElement.scroll({
+            left: Math.max(0, width / 2 - 500)
```
XanderVertegaal, line 35: Is 500 here chosen as an arbitrary (but sensible) minimal width?

frontend/src/app/services/parse.service.ts
```diff
@@ -21,6 +23,142 @@ export class ParseService {
...
+    const todo = [[nltk, tree]];
+    // The core of the trampoline. Process one thunk at a time. We know we are
+    // done when the list is empty.
+    while (todo.length) {
+        // The algorithm is written such that the order in which we process
+        // child nodes does not matter, so we can simply pop thunks off the back
+        // of the todo list. tsc is apparently unable to infer that `todo` is
+        // guaranteed to be nonempty on the next line, hence the silencing
+        // comment.
+        // @ts-ignore
```
XanderVertegaal: I understand why the `@ts-ignore` is here, but instead of silencing it completely, maybe a type assertion `todo.pop() as [any, any]` or a non-null assertion `todo.pop()!` would be neater.

This is not something I insist on, but rather something you might want to keep in mind when you add the types.

frontend/src/app/services/parse.service.ts
```diff
@@ -21,6 +23,142 @@ export class ParseService {
...
+        let task: [any, any] = todo.pop();
```
XanderVertegaal: Why are these `let`s rather than `const`s?

frontend/src/app/services/parse.service.ts
```diff
@@ -21,6 +23,142 @@ export class ParseService {
...
+ * stack overflow.
+ */
+export function nltk2tableau(nltk: any) {
+    const tree = emptyTableau();
+    // We will be using trampolining instead of recursion. A general but
+    // admittedly not very helpful description can be found in the first bullet
+    // of
+    // https://en.wikipedia.org/wiki/Trampoline_(computing)#High-level_programming.
+    // The gist is this: instead of recursing, functions return pieces of work
+    // ("thunks") that should be executed next. The trampoline, which in this
```
XanderVertegaal, line 97: Interesting. Would you say trampolining is generally superior to recursion when building nested datastructures like this? Are there any trade-offs, or areas in which recursion is preferable?

frontend/src/app/services/parse.service.ts
```diff
@@ -21,6 +23,142 @@ export class ParseService {
...
+    // Our own format does not follow this rule: chains of nodes with single
+    // children are stored as adjacent elements in an array. `tree` is the
+    // object that contains this array (`tree.nodes`). When there is a
+    // bifurcation, the subtrees are stored inside `tree` rather than in the
+    // bifurcating node.
+    const node: any = {};
+    tree.nodes.push(node);
+    // Tease out the different parts of the stringified node payload. The
+    // following lines will be rewritten using regular expressions.
+    const lines = nltk.node.split('\n');
```
XanderVertegaal: Perhaps this string decoding could live in its own function (so we can test it in isolation), but you might want to wait for that until you have introduced the regexes.

XanderVertegaal (commented):
Very neat, thank you for the documentation! Have a look at the alignment of the circles+symbols at the end of the trees. Otherwise, please carry on!

frontend/src/app/annotate/tableau-svg/tableau-node.component.svg
jgonggrijp: Argh! Which browser is this?

frontend/src/app/annotate/tableau-svg/tableau-svg.component.ts
jgonggrijp: It's deliberately slightly more than half of the constrained body width. Without it, the center of the root would be on the left edge. I went with a slight bias towards the right because the description of the node starts on the left.

frontend/src/app/services/parse.service.ts
jgonggrijp: Because I'm still used to `var` semantics and mentally hoisting them to the top of the function scope. 😅 (To be honest, I don't believe that `let` and `const` were an improvement to the language and I'm only using them in order to respect the local code convention.) I'll change it to `const`.

frontend/src/app/services/parse.service.ts
jgonggrijp: When the datastructure might be arbitrarily deeply nested, you kind of "have" to use trampolining in order to avoid stack overflows. I became aware of this because of https://github.com/jashkenas/underscore/security/advisories/GHSA-qpx9-hpmf-5gmw and now cannot un-see that potential problem anymore, hence my vigilance. I'm actually planning to add a generalized `_.trampoline` to Underscore in order to provide a foundation for solutions like this one.

Suprisingly, there is no significant cost to trampolining versus recursion performancewise. The number of function calls is the same and the memory consumption is similar.

Conceptually, trampolining is one possible way to implement continuation passing style. Which has some theoretical elegance, but is more taxing on the brain than the conventional "direct" style. I would say that recursion is still preferable when you can get away with it, but in this case I went straight for trampolining because I think there might actually be a real risk that these trees get (slightly) too large.

jgonggrijp:
> Your plan corresponds well to what we agreed, so please go ahead with the implementation. The only item on the list I would like you to clarify is 10: what do you mean with the 'no trigger unless viewing parse tab'?

The effect that I described to you previously, where the tableau would not render unless the "CCG / LLF" tab was selected while I pressed the "Parse and prove" button. However, this seems to have been solved already, possibly because I accepted your suggestion to add the `share` operator. It does not matter anymore which tab is shown when I press the button.

I'll double-check after rebasing just to make sure, though.

frontend/src/app/annotate/tableau-svg/tableau-node.component.svg
jgonggrijp: By the way, I investigated using `faLock` and `faLockOpen` instead, but the icons would display with zero width and become invisible. I wasn't able to make it work and I suspect it is because the SVG icon is nested in an HTML element which is then again nested inside the larger SVG tableau.

frontend/src/app/annotate/tableau-svg/tableau-svg.component.ts
jgonggrijp: You seem to have been right that the root is always at the middle of the top edge, by the way.

frontend/src/app/annotate/tableau-svg/tableau-node.component.svg
jgonggrijp: > Argh! Which browser is this?

Never mind, it happens in my browser, too. Apparently I had to rebuild the container to see it.

jgonggrijp:
@XanderVertegaal I force-pushed again because I needed to insert some of the new types early in history. Otherwise, all of the intermediate commits would be uncompilable.

I did `docker compose down` in order to rebase and then did `docker compose --build up` afterwards to check that all the tests would still pass. They did, but when trying to visit the frontend, I discovered that my templates don't typecheck yet. I will fix that later but I need to leave the library now. You are welcome to review the types in the meanwhile.

The only changes I still expect to make is type fixes in the templates and/or a simplification in the `@/types`, where I simply collapse all `ProofNode` subtypes into a single object type with all fields optional. That is a bit less correct but easier to work with. In the latter case I might rebase the last two commits one more time.

frontend/src/app/annotate/tableau-svg/tableau-node.component.svg
XanderVertegaal: I see. For what it is worth: I am using Firefox.

jgonggrijp:
Done.

frontend/src/app/annotate/annotation-menu/annotation-menu.component.html
```diff
@@ -1,35 +1,54 @@
...
+            </section>
+        </ng-template>
+    </li>
+    <li [ngbNavItem]="3">
+        <button ngbNavLink>
+            <b class="me-2" aria-hidden>⊥</b>
+            <span i18n>Proof / Contradiction</span>
+        </button>
+        <ng-template ngbNavContent>
+            <section style="overflow-x: scroll;">
```
XanderVertegaal:
```suggestion
            <section class="overflow-x-scroll">
```

frontend/src/app/annotate/tableau-svg/tableau-svg.component.ts
```diff
@@ -25,59 +28,11 @@ export class TableauSVG {
...
-                    {idx: 21, term: 'expensive@car', label: 'c2', rule: 'up_mon_fun[19,20]', bg: 'honeydew'},
-                    {idx: 22, term: 'car', label: 'c2', rule: 'up_mon_fun[19,20]', bg: 'mistyrose'},
-                    {idx: 23, term: 'car', label: 'c2', rule:'int_mod_tr[21]', bg: 'honeydew'},
-                    {idx: 24, term: 'expensive', label: 'c2', rule:'int_mod_tr[21]', bg: 'honeydew'},
-                    {end: true, rule: 'cl_subsmption[23,22]'}
-                ]
-            }
-        ]
-    }
+    public readonly tree = input.required<ProofTree>();
```
XanderVertegaal: (Utterly minor: I think `input`s are [conventionally](https://angular.dev/style-guide#group-angular-specific-properties-before-methods) put at the top of the component. At least, that is where most people will expect to find them.)

frontend/src/app/annotate/tableau-svg/tableau-tree.component.ts
```diff
@@ -1,32 +1,32 @@
...
-    _tree: any;
+    _tree: ProofTree = {nodes: []};
     @Input()
-    get tree(): any {
+    get tree(): ProofTree {
         return this._tree;
     }
 
-    set tree(value: any) {
+    set tree(value: ProofTree) {
```
XanderVertegaal, line 22: It feels so good to see these `any`s replaced with actual types <3

frontend/src/app/services/parse.service.ts
```diff
@@ -31,11 +33,154 @@ export class ParseService {
...
+    // The core of the trampoline. Process one thunk at a time. We know we are
+    // done when the list is empty.
+    while (todo.length) {
+        // The algorithm is written such that the order in which we process
+        // child nodes does not matter, so we can simply pop thunks off the back
+        // of the todo list. tsc is apparently unable to infer that `todo` is
+        // guaranteed to be nonempty on the next line, hence the silencing
+        // comment.
+        // @ts-ignore
+        const task = todo.pop() as ProofConversionThunk;
```
XanderVertegaal: With your type assertion in place, the silencing comment can go.

```suggestion
        // guaranteed to be nonempty on the next line; hence
        // the type assertion.
        const task = todo.pop() as ProofConversionThunk;
```

(I would use `, whence` or `; hence`, instead of `, hence`, but leave it if you prefer.)

XanderVertegaal (approved):
Great work! 👍

XanderVertegaal:
image

Glorious! 😍

PR #97, opened 2026-04-21
Feature/parse and prove request status

XanderVertegaal:
Two small quality of life changes:

- The Parse and Prove button now shows a loading indicator while the request is in flight.
- Toasts are shown when a parse fails (e.g. if the LangPro container is down).


jgonggrijp (approved):
Nice, very clean. 👍

PR #98, opened 2026-04-21
Add NoData component

### LINKED ISSUE #82: Unified "No parse/proof data available yet"
Before 'Parse and Prove' is clicked, the tabs where the proof trees and parse tables will appear are empty. As per #78 there is a message with some help text, but ideally this should be outfactored and used in both tabs.

A new component is perhaps a bit too much. Slots/transclusion is also an option.

XanderVertegaal:
Closes #82


jgonggrijp (approved):
Nice and simple. 👍

PR #99, opened 2026-04-21
Add missing Jasmine types to TS config

XanderVertegaal:
Somehow, my IDE has gotten very unhappy without these types, and I do not know what caused it. Does your IDE have the same issue?

image

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -48,6 +48,7 @@
     "@angular/localize": "19.2.13",
     "@types/express": "^4.17.17",
     "@types/jasmine": "~5.1.0",
+    "@types/jest": "^30.0.0",
     "@types/node": "^18.18.0",
     "jasmine-core": "~5.1.0",
     "karma": "~6.4.0",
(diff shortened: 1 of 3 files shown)


jgonggrijp (changes requested):
I can try whether my IDE complains about it next Wednesday (not scheduled for CDH work today), but I suspect it doesn't because I usually don't configure it to do live typechecking.

Regardless, adding types for `jest` is probably not the right solution because the tests are based on `jasmine`. We already have `@types/jasmine` in the dependencies, but maybe it also needs to be listed in the `tsconfig` as you did for `jest`. Otherwise, we might be looking at a bug that was introduced with a recently updated dependency?

XanderVertegaal:
I tried to delete a reply I made and accidentally deleted the whole comment - my apologies.

For the record: your suggestion to replace the `jest` types with `jasmine` works as well. I assumed that our migration to Angular 19 also included the migration to Jest, but we're still on Jasmine indeed.

PR #100, opened 2026-04-24
More compact parse tables

### LINKED ISSUE #49: Implement rendering of CCG parser results

### LINKED ISSUE #81: Reduce vertical space in syntactic parse tables
Reduce vertical margins/padding and put rule + content on the same line. Differential styling is nice if it can be maintained, but this is not a must.

XanderVertegaal:
Closes #49 

This branch was developed using changes from [this PR](https://github.com/CentreForDigitalHumanities/langpro-container/pull/8) and [this PR](https://github.com/CentreForDigitalHumanities/langpro-container/pull/4) on `langpro-container`. Those two PRs should be merged before this PR can be reviewed. I have developed this PR based on a local branch where I've integrated the changes from both PRs.

This PR enables the final three CCG parse artifacts (i.e. the CCG Term, CCG Corrected Term and LLF). In a meeting with @kovvalsky, the following guidelines for styling the nodes were laid out.

- The chunker output (e.g. "I-NP") in leaf nodes can be omitted.
- For each node, the server returns a string consisting of a "rule" and "content", in the format `rule[content]`. These brackets surrounding are to be removed from the output.
- Rules may be formatted in a grey box, as long as they are put on the same line as the content (to avoid the table from becoming taller than it needs to be).
- The rule `@` is common and trivial and should be filtered out.
- A colon `:` marks the start of a 'feature', which is set in subscript. The end of a string or a special character `()[]/\-` terminates the feature.
- Node content should appear in uppercase, but only in "CCG Tree" output, except for those parts of the content that are a 'feature' or are found in a list of whitelisted elements (i.e. 'period' or 'conj').
- Node content should not wrap across multiple lines.

image


frontend/src/app/annotate/annotation-parse-results/parse-tree-table/parse-tree-table.component.spec.ts
```diff
@@ -35,3 +35,46 @@ describe('ParseTreeTableComponent', () => {
...
+
+        it('should strip multiple extra brackets from content', () => {
+            const result = extractRule('fa[s:[ng]-[np]]');
+            expect(result).toEqual({ rule: 'fa', content: 's:ng-np' });
+        });
+
+        it('should strip all internal brackets', () => {
+            const result = extractRule('ba[[[s:dcl]]]');
+            expect(result).toEqual({ rule: 'ba', content: 's:dcl' });
+        });
```
jgonggrijp, line 71: I'm not against having tests for this, or against having a forgiving parser, but it is a little bit surprising that the annotator needs to account for this. Can't the container just send well-formed notations?

frontend/src/app/annotate/annotation-parse-results/parse-tree-table/parse-tree-table.component.ts
```diff
@@ -63,42 +63,42 @@ function buildUnaryNode(node: UnaryNode): TreeNodeDisplay {
...
+        rule,
         children: [child]
     };
 }
 
 /**
  * Parses a node string to extract the rule and the content.
  *
- * A node string is usually of the form "A(B)", where a is the rule applied
+ * A node string is usually of the form "A[B]", where a is the rule applied
```
jgonggrijp:
```suggestion
 * A node string is usually of the form "A[B]", where A is the rule applied
```

frontend/src/app/annotate/annotation-parse-results/parse-tree-table/parse-tree-table.component.ts
```diff
@@ -63,42 +63,42 @@ function buildUnaryNode(node: UnaryNode): TreeNodeDisplay {
...
-    const lastParen = nodeString.lastIndexOf(')');
+export function extractRule(nodeString: string): { rule: string, content: string; } {
+    const firstBracket = nodeString.indexOf('[');
+    const lastBracket = nodeString.lastIndexOf(']');
 
     // Return a fallback value if the string is not what we expect.
-    if (firstParen === -1 || lastParen === -1 || lastParen < firstParen) {
+    if (firstBracket === -1 || lastBracket === -1 || lastBracket < firstBracket) {
         return {
             rule: "",
```
jgonggrijp: Perhaps it would be a bit cleaner to still acknowledge the `@` rule in the internal data model, and use a separate `rulePresentation` field that can be the empty string for `@`. I don't care enough about it to make this a condition for approval, though.

frontend/src/app/annotate/annotation-parse-results/parse-tree-table/tree-node.component.html
jgonggrijp, line 1: This template makes lots of `node()` calls. Maybe that is worth dedicating an `@let` to?

frontend/src/app/annotate/annotation-parse-results/parse-tree-table/parse-tree-table.component.ts
jgonggrijp: Come to think of it, you might even just leave `@` in the internal presentation and address the presentation with a pipe instead.

frontend/src/app/pipes/subscript-pipe.spec.ts
jgonggrijp, line 1: Love the thoroughness of these tests.

frontend/src/app/pipes/subscript-pipe.ts
```diff
@@ -0,0 +1,21 @@
...
+     * Transforms text by converting all content in between a colon and a
+     * slash, a parenthesis, a hyphen or the end of the string to subscript.
+     * Example: "np:dcl\nb" becomes "np<sub>dcl</sub>nb"
+     */
+    transform(value: string): string {
+        if (!value) {
+            return value;
+        }
+
+        return value.replace(/:(.*?)(\\|\/|$|\)|\(|-)/g, (_, p1, p2) => `<sub>${p1.toLocaleLowerCase()}</sub>${p2}`);
```
jgonggrijp: The arrow function here can be a constant at module scope. Saves recreating it every time the `transform` method runs.

frontend/src/app/annotate/annotation-parse-results/parse-tree-table/tree-node.component.html
XanderVertegaal: Done!

PR #107, opened 2026-05-14
Remove unused problem list view

XanderVertegaal:
This removes the unused `list()` method from `ProblemView`, probably from an earlier stage of development where we considered having a list. Everything seems to work perfectly without it.

We even had a few tests for a 'list' view in `TestProblemViewPermissions`. These keep working even after removing `.list()`, since `ModelViewSet` provides a default list method that simply returns the whole queryset. That is going to be a problem (see what I did there?) with the new 'hidden' feature on problems (#106).

To simplify things, I propose overriding the default list method and returning 404 NOT FOUND. The tests for the list view can then also go.

### DIFF
--- a/backend/problem/views/problem.py
+++ b/backend/problem/views/problem.py
@@ -1,5 +1,4 @@
 from rest_framework.decorators import action
-from rest_framework.exceptions import MethodNotAllowed
 from rest_framework.request import Request
 from rest_framework.response import Response
 from rest_framework.viewsets import ModelViewSet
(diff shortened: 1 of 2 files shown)


backend/problem/views/problem.py
```diff
@@ -51,18 +52,7 @@ def get_permissions(self):
...
     def list(self, request: Request) -> Response:
-        """
-        Lists all Problems in the database, with optional filtering.
-        """
-        filters = get_filters(request.query_params)
-
-        qs = self.get_queryset()
-
-        if filters is not None:
-            qs = qs.filter(filters)
```
jgonggrijp, line 62: You could also just filter `qs` by `hidden=False` if the current user lacks appropriate permissions (after merging #106). I don't have a strong opinion on the matter though. We can always revert this later if we feel we need a list endpoint.

jgonggrijp (approved):

PR #109, opened 2026-05-25
Feature/annotation log

### LINKED ISSUE #69: Annotation log
Each problem should have a list showing the history of annotations.

XanderVertegaal:
Closes #69

This PR adds a simple list of annotations to each problem, so users can see when annotations were added/removed and by whom.

image

The annotations are sorted by date from most to least recent. As a bonus, I've added some frontend filtering. We can always add backend filtering, sorting, pagination and other fancy list goodies if/when they are requested.


frontend/src/app/annotate/annotation-comments/annotation-comments.component.html
```diff
@@ -1,4 +1,112 @@
...
+</div>
+@if (events.length === 0) {
+<p class="text-muted small fst-italic" i18n>
+    No annotations match the current filters.
+</p>
+} @else {
+<ul class="list-unstyled d-flex flex-column gap-2 mb-0">
+    @for (event of events; track $index) {
+    @let added = event.eventKind === "added";
+    @let isLabel = event.annotationKind === "label";
```
jgonggrijp: I think the `@for` and `@let` statements deserve to be at different indentation levels. Maybe keep the `@for` with the `<ul>` above? What do the conventions say?

frontend/src/app/annotate/annotation-comments/annotation-comments.component.ts
```diff
@@ -1,10 +1,104 @@
...
+        events.push({
+            eventKind: "removed",
+            annotationKind,
+            timestamp: annotation.removedAt,
+            actor: annotation.removedBy,
+            description,
+        });
+    }
+    return events;
+}
```
jgonggrijp, line 53: This is model-level logic. Maybe it deserves living in a dedicated module?

frontend/src/app/annotate/annotation-comments/annotation-comments.component.ts
```diff
@@ -1,10 +1,104 @@
...
+
+    public filteredEvents = computed(() => {
+        const events = this.allEvents();
+        if (events === null) {
+            return null;
+        }
+        return events.filter(
+            (e) =>
+                (e.eventKind === "added" ? this.showAdded() : this.showRemoved()) &&
+                (e.annotationKind === "label" ? this.showLabels() : this.showKB())
```
jgonggrijp, line 93: Does this mean that you can either see all events, **or** filter by event kind and annotation kind at the same time, but **not** filter only by event kind or only by annotation kind?

jgonggrijp (commented):
I have reviewed the code and the screenshot (which looks nice!), but did not try the changes locally. Based on the code, I wonder whether it is possible to see both additions and removals but only for labels, or only removals but both for labels and KB items. Other than that, the functionality looks very nice.

frontend/src/app/annotate/annotation-comments/annotation-comments.component.html
XanderVertegaal: Angular seems to have changed its conventions. It used to be the case that structural directives like `@if` and `@for` did not change the indentation, but now they do (and I'm happy about it)! I'll change it.

frontend/src/app/annotate/annotation-comments/annotation-comments.component.ts
XanderVertegaal: I interpret this as data-to-representation conversion logic and would say that it therefore belongs close to the component that will eventually render it. I would not think putting this code in a dedicated module makes sense unless it is going to be shared across components, which I think is unlikely. Having said this, I am happy to discuss this and be convinced otherwise! We seem to have different opinions on where to put code and I am interested in your considerations and thoughts about this.

frontend/src/app/annotate/annotation-comments/annotation-comments.component.ts
XanderVertegaal: The four buttons 'Added', 'Removed', 'Labels' and 'KB items' can independently be toggled on and off, so you have full flexibility. For instance, you can select:

- labels + added (= all added labels)
- labels + added + removed (= all labels)
- labels + KB items + added + removed (= everything)

Obviously, some combinations are not very useful:
- added + removed (= nothing, because you haven't selected labels and/or KB items)
- labels + KB items (= nothing, because you haven't selected added and/or removed)
- nothing!

frontend/src/app/annotate/annotation-comments/annotation-comments.component.ts
jgonggrijp: Let's chat a bit about this when we see each other, then. No need to keep the PR on hold just for that chat, by the way.

XanderVertegaal: Yes, let's! In that case I'll go ahead and merge after fixing the failing unit tests.

